PR #351, opened 2016-02-19
ANFOG_RT/DM pipelines

danfruehauf:
Scripts to process bundle of files uploaded as zip archive for
both ANFOG_RT and ANFOG_DM.


danfruehauf:
@bpasquer Added another commit which refactors things a bit. There is a new function `nc_list_variables` which then we can `grep` on to find variables by regex.

ANFOG/DM/anfog_dm_netcdf_compliance.sh
```diff
@@ -0,0 +1,146 @@
...
+        if [ x"$var" != x ]; then
+            nc_set_att -a long_name,$var,o,c,"${var,,} informations" $nc_file
+        fi
+    done
+}
+
+fix_depth_min_max() {
+    nc_file=$1; shift
+    # val.nc WTF?!
+    local depth_min=`ncap2 -O -C -v -s "val=DEPTH.min();print(val)" ${nc_file} ~/val.nc | cut -f 3- -d ' ' ;`
```
danfruehauf: @bpasquer I'm not sure what `~/val.nc` should refer to.

bpasquer: this is a compulsory argument -output- of the ncap2 function.  I should add a line to delete straight after extracting the min max value. we can rename it 'tmp' if you prefer

danfruehauf: Perhaps you should then:
 * Create a temporary file
 * Output to it
 * Delete it

danfruehauf: Or you can also output to `/dev/null`.

PR #372, opened 2016-03-06
AATAMS_SATTAG_DM ignore TDR dir in lftp sync

danfruehauf:
Contains txt files which we should not harvest


danfruehauf:
FYI I also cleaned the WIP dirs from those txt and zip files.

xhoenner:
Excellent, thanks @danfruehauf!

PR #373, opened 2016-03-07
Remove opendap and public from doc

danfruehauf:
Usage of opendap and public is not permitted any more. Also
point archive dir to the right place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ A place to add Data Services scripts from PO's
 
 # Folder stucture
 
-The suggested naming convention we agreed on with the developpers, regarding the different PO's scripts was :
+The suggested naming convention we agreed on with the developers, regarding the different PO's scripts was :
 *[FACILITY_NAME]/[SUB-FACILITY_NAME]_[script_name]*
 
 example :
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,18 +32,13 @@ The environment variables are:
...
 export INCOMING_DIR='/tmp/incoming'
-export OPENDAP_IMOS_DIR="$OPENDAP_DIR/1/IMOS/opendap"
-export PUBLIC_IMOS_DIR="$PUBLIC_DIR"
-export ARCHIVE_IMOS_DIR="$ARCHIVE_DIR"
 export WIP_DIR='/tmp/wip'
 export DATA_SERVICES_DIR="$PWD"
 export LOG_DIR='/tmp/log'
 
-mkdir -p $OPENDAP_IMOS_DIR $PUBLIC_IMOS_DIR $ARCHIVE_IMOS_DIR $WIP_DIR $LOG_DIR
+mkdir -p $ARCHIVE_DIR $WIP_DIR $LOG_DIR
```
mhidas: Do we still need this? Testing now happens in the VM, where the env is set up with chef.
If we do want to leave it, should also include `mkdir  $INCOMING_DIR`

danfruehauf: I think it is redundant, as the VM takes care of this. In general, most of this env section in the README is redundant. But @lbesnard wanted it.

The po box should take care of all of this AFAIK.

mhidas: So in that case, for completeness, the above line should also include creating `$INCOMING_DIR`

danfruehauf:
@mhidas I'll talk to @lbesnard about what should be in that README.

I've added the creation of `$INCOMING_DIR`.

PR #374, opened 2016-03-08
Refactor ANMN dest_path.py and FileClassifier class

mhidas:
No change in functionality, just moving more generic code into `lib/python` so that ABOS pipeline can make use of it.

* move more generic methods from ANMNFileClassifier into FileClassifier base class
* rename ANMNFileClassifier to MooringFileClassifier and move to lib/python
* update tests (no longer need ANMN test_dest_path)


lib/python/file_classifier.py
```diff
@@ -38,6 +40,33 @@ def _error(cls, message):
         "Raise an exception with the given message."
         raise FileClassifierException, message
 
+
+    @classmethod
+    def _get_file_name_fields(cls, input_file, min_fields=6):
+        """Return the '_'-separated fields in the file name as a list.
+        Raise an exception if the number of fields is less than min_fields.
+        """
+        just_the_name = re.sub('\.nc$', '', os.path.basename(input_file))
```
danfruehauf: I don't know if you want to note it, but this function will work only on files ending with `nc`. And if it doesn't? Unexpected behaviour?

lib/python/file_classifier.py
```diff
@@ -90,3 +130,99 @@ def _make_path(cls, dir_list):
         for i in range(len(dir_list)):
             dir_list[i] = str(dir_list[i])
         return os.path.join(*dir_list)
+
```
danfruehauf: Extra newline here.

lib/python/file_classifier.py
```diff
@@ -90,3 +130,99 @@ def _make_path(cls, dir_list):
...
+        CTD_timeseires, Biogeochem_profile, etc..)
+
+        """
+
+        feature_type = cls._get_nc_att(input_file, 'featureType')
+        if feature_type == 'profile':
+            return 'Biogeochem_profiles'
+
+        var_names = set(cls._get_variable_names(input_file))
+        salinity = set(['PSAL', 'CNDC'])
```
danfruehauf: Those variables can probably become a hash. And then you will not need a lot of `if`s down below, but something to iterate on the hash.

danfruehauf:
Looks alright to me. Some minor comments, nothing big.

lib/python/file_classifier.py
mhidas: I like to have a bit of extra space to separate the two classes.

danfruehauf: 2 lines is the standard.

danfruehauf: https://www.python.org/dev/peps/pep-0008/#blank-lines

lib/python/file_classifier.py
mhidas: Yes, and if there were many more categories that would be preferable. As it is, I think the logic is clearer this way. Also, the order these tests are applied is important.

danfruehauf: That's OK, so leave it as is.

lib/python/file_classifier.py
mhidas: Ok.

mhidas:
Thanks @danfruehauf I've fixed the other two things

PR #382, opened 2016-03-17
SOOP - cleaning

lbesnard:
* follow conventions
 * use ship_callsign library
 * no need to change watch.d scripts


lbesnard:
@bpasquer you might be interested by this. using the ship_callsign lib so all boats are in one file

danfruehauf:
I'll give my :+1: here and let @bpasquer also have a look and potentially merge.

bpasquer:
:+1:

PR #385, opened 2016-03-18
SRS_ALTIMETRY trigger

danfruehauf:


danfruehauf:
Can be merged. Uses `/bin/false` as trigger. At least it'll put all the rest of the things in place (chef wise).

PR #388, opened 2016-03-21
ANFOG_DM pipeline :fix function for IMOS compliance of QC var

bpasquer:


bpasquer:
@danfruehauf  That'll solve part of our problem

PR #394, opened 2016-03-23
WIP python impl of pipeline functions

danfruehauf:
**DO NOT MERGE - REUSE, CUT, PASTE**

One of the mistakes done was choosing Bash as the main pipeline language, this is due to me believing most use cases in the data pipeline will be quite simple + assuming we will need to use Bash/Python/Matlab. Little did I know.

This is a beginning of an implementation of some of the Bash functionality in python (push files, trigger harvesters, etc). There's heaps that's still missing, but someone can probably iterate on that work, or grab snippets from here that can be used in future work if the pipeline will be completely ported to python.

### DIFF
--- a/SOOP/SOOP_CO2/incoming_handler.py
+++ b/SOOP/SOOP_CO2/incoming_handler.py
@@ -1,0 +1,106 @@
+#!/usr/bin/python
+
+import os
+import sys
+import tempfile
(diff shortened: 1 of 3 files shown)


mhidas:
I don't think we need this anymore as we're taking a different approach.

PR #401, opened 2016-03-30
ANMN pipeline improvements

mhidas:
* Handle PDF logsheets, CNV files, and PNG plots from Toolbox.
* Unit tests for non-netcdf files.
* Use `$TMPDIR` for unzipping files & clean them up at end.
* Add trigger checkers, selected in watch.d script (none by default).

### DIFF
--- a/ANMN/common/incoming_handler.sh
+++ b/ANMN/common/incoming_handler.sh
@@ -3,8 +3,6 @@
 export PYTHONPATH="$DATA_SERVICES_DIR/lib/python"
 export SCRIPTPATH="$DATA_SERVICES_DIR/ANMN/common"
 
-UNZIP_DIR="$WIP_DIR/$JOB_NAME"
-
 declare -r BACKUP_RECIPIENT=marty.hidas@utas.edu.au
 declare -r DATACODE="[A-Z]+"
 declare -r TIMESTAMP="[0-9]{8}T[0-9]{6}Z"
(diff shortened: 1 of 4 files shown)


mhidas:
@bpasquer @xhoenner @lbesnard Let me know if you have time for a quick look at this today. Otherwise I'll just merge it myself. I would like to deploy it this afternoon.

bpasquer:
i can have a look after lunch

ANMN/common/incoming_handler.sh
```diff
@@ -60,11 +58,13 @@ handle_netcdf() {
...
+    chmod 755 $unzip_dir  # make directory world readable
+
     # unzip the file
-    log_info "Unzipping '$zipfile' to $UNZIP_DIR"
-    mkdir -pv $UNZIP_DIR
-    local extracted_files=`mktemp`
-    unzip_file $zipfile $UNZIP_DIR $extracted_files || file_error "Failed to unzip"
+    log_info "Unzipping '$zipfile' to $unzip_dir"
+    local extracted_files=`mktemp --tmpdir=$unzip_dir`
+    unzip_file $zipfile $unzip_dir $extracted_files || file_error "Failed to unzip"
```
bpasquer: just for the ease of understanding you could rename the variable extracted_files into something that reflect better the fact that is a manifest file

bpasquer:
good to go @mhidas  :+1:

ANMN/common/incoming_handler.sh
mhidas: Thanks for the comment, but I'm not sure what you mean. So far I've only heard the term "manifest file" used in the context of bulk indexing, and that is not what this is. It's simply a list of files that were extracted from the zipfile. So I think `extracted_files` is a pretty good name for it. I could call it `extracted_files_list` if that makes it clearer?

bpasquer:
well it is a manifest in a sense that it is a document that give all the details about the content of the zip . But extracted_files_list works prefectly well too :+1:

mhidas:
@bpasquer Thanks. I'll settle for that.

PR #417, opened 2016-04-18
ANFOG - remove tmpdir

lbesnard:
* tmpdir were not removed in the same way in script

### DIFF
--- a/ANFOG/DM/incoming_handler.sh
+++ b/ANFOG/DM/incoming_handler.sh
@@ -98,14 +98,14 @@ handle_zip_file() {
 
     unzip_file $file $tmp_dir $tmp_zip_manifest
     if [ $? -ne 0 ]; then
-        rmdir $tmp_dir
+        rm -rf --preserve-root $tmp_dir
         file_error "Error unzipping"
     fi
 
     local nc_file
     nc_file=`grep ".*_FV01_.*\.nc" $tmp_zip_manifest | head -1`
     if [ $? -ne 0 ]; then
-        rmdir $tmp_dir
+        rm -rf --preserve-root $tmp_dir
         file_error "Cannot find NetCDF file in zip bundle"
     fi
 


lbesnard:
@bpasquer

julian1:
How do you know the leaking of resources are not symptomatic of a more general bug/design issue?

julian1:
It looks like the code is written with an assumption that if mktemp fails, then it will not extract anything into the tmp directory. But given that it's not cleaning up properly this is not always the case. 

Are we sure that we shouldn't be checking return value of mktemp needs to be interpreted more carefully? I ask because "rm -rf" is such a dangerous action.

julian1:
I would really like to know what the condition is where mkdir unzips the files but still reports an error. Maybe the return value is something trivial, like a bad date encoding, or permission or something.

bpasquer:
The code is written with the assumption that unzip might fail, not mktemp. For example  it happens when the zip archive is corrupted. 
As for the rm -rf  , the option --preserve root is used because it is dangerous indeed, but as far as i understand it is the only way to remove directories recursively.

julian1:
You're correct, that I meant unzip not mktemp. 

I really think there should be more fine-grained handling for non-zero responses rather than blindly deleting the output.

From the unzip man page,

```
             0      normal; no errors or warnings detected.

              1      one or more warning errors were encountered, but processing completed successfully anyway.  This includes zipfiles where one or more files was skipped due to unsupported compression method or encryption  with  an
                     unknown password.

              2      a generic error in the zipfile format was detected.  Processing may have completed successfully anyway; some broken zipfiles created by other archivers have simple work-arounds.

              3      a severe error in the zipfile format was detected.  Processing probably failed immediately.
```

For example, I'm not sure it makes sense to delete all the files, when processing completes successfully - simply because some other archiver may not handle it.

lbesnard:
@julian1 , I think that if there is an issue while unzipping a file for any obscure reason, it will just go into the ERROR_DIR, and then @bpasquer will investigate why.

bpasquer:
@julian exit status 0 is the only acceptable outcome when it comes to unzipping. We didn't design the pipeline to handle errors. And like @lbesnard said, when an problem is detected the zip goes in the error dir anyway, nothing gets lost.

julian1:
Ok, I didn't realize that the file would be placed in ERROR_DIR

bpasquer:
No worries , i appreciate you had a look at what we did and commented

PR #422, opened 2016-04-19
Fix #410

### LINKED ISSUE #410: File disappears after going to $INCOMING_DIR
Some files completely disappear from the system after going to $INCOMING_DIR

# How to reproduce
on aws10, connect as PO user, and do a ```source ./env``` at the base of data-services dir (would be good to add this automatically via chef btw)

```bash
# copy a SOOP ASF MT file to its incoming dir
cp $WIP_DIR/SOOP/SOOP_XBT_ASF_SST/data_unsorted/ASF_SST/ship/2016/IMOS_SOOP-ASF_MT_20160404T000000Z_VLMJ_FV01_C-20160405T040053Z.nc $INCOMING_DIR/SOOP/ASF/MT   

input_logf SOOP_ASF_MT
Apr  6 14:44:36 10-aws-syd SOOP_ASF_MT: '/mnt/ebs/tmp/tmp.oltFjHYb6U/IMOS_SOOP-ASF_MT_20160404T000000Z_VLMJ_FV01_C-20160405T040053Z.nc' is not a valid file, aborting.

### similar behaviour with 
cp $WIP_DIR/SOOP/SOOP_XBT_ASF_SST/data_unsorted/ASF_SST/ship/2016/IMOS_SOOP-SST_T_20160405T000300Z_VRDU8_FV01_C-20160406T001307Z.nc $INCOMING_DIR/SOOP/SST

input_logf SOOP_SST
Apr  6 15:16:27 10-aws-syd SOOP_SST: '/mnt/ebs/tmp/tmp.hr9huOeRry/IMOS_SOOP-SST_T_20160405T000300Z_VRDU8_FV01_C-20160406T001307Z.nc' is not a valid file, aborting.
```

Ok why not, the file is 'not valid'. Well the original one is. But it looks like there is an issue here https://github.com/aodn/data-services/blob/master/lib/common/util.sh#L165 with the tmp file. The definition of 'valid' is a bash check to see if the file exists or not. So we're really at a low level check.

The issue is that we should still be able to locate the 'bad' file. But it is nor in $ERROR_DIR nor in $INCOMING_DIR. Anyway, this is **EXTREMELY** concerning. Of course, Nagios won't warn us on anything because there is no files in the ERROR_DIR. So we have potentially lost many files.

Both these folders are empty, as well as the thredds folder where the file should go to.
```
ls -la $INCOMING_DIR/SOOP/ASF/MT 
ls -la $ERROR_DIR/SOOP_ASF_MT
```

I wish I was wrong, I've tried and checked many times. 

@smancini @jonescc @pblain

julian1:
Netcdf file that would silently disappear now routed to the error directory, 
```
root@10-aws-syd:/mnt/ebs/log/data-services$ rm /mnt/ebs/error/SOOP_SST/*
root@10-aws-syd:/mnt/ebs/log/data-services$ ls /mnt/ebs/error/SOOP_SST/
root@10-aws-syd:/mnt/ebs/log/data-services$ cp /mnt/ebs/wip/SOOP/SOOP_XBT_ASF_SST/data_unsorted/ASF_SST/ship/2016/IMOS_SOOP-SST_T_20160223T000300Z_VRDU8_FV01_C-20160224T001314Z.nc $INCOMING_DIR/SOOP/SST
root@10-aws-syd:/mnt/ebs/log/data-services$ ls /mnt/ebs/error/SOOP_SST/                                                                     
IMOS_SOOP-SST_T_20160223T000300Z_VRDU8_FV01_C-20160224T001314Z.nc.20160420-092036
root@10-aws-syd:/mnt/ebs/log/data-services$ 
```

### DIFF
--- a/bin/generic_incoming_handler.sh
+++ b/bin/generic_incoming_handler.sh
@@ -71,7 +71,15 @@ main() {
     local path_hierarchy
     path_hierarchy=`$DATA_SERVICES_DIR/$path_evaluation_executable $file`
     if [ $? -ne 0 ] || [ x"$path_hierarchy" = x ]; then
-        rm -f $tmp_file
+
+        # if the trigger_checkers_and_add_signature() does not return a modified filename then do
+        # not delete $tmp_file which refers to the same resource as $file.
+        # this permits the caller to correctly handle $file as an unprocessed error
+        if [ "$tmp_file" != "$file" ]; then
+            rm -f $tmp_file
+        fi
+
+        # Note, file_error calls exit() which will short-circuit the call to S3_put()
         file_error "Could not evaluate path for '$file' using '$path_evaluation_executable'"
     fi
 


mhidas:
@julian1 Nice work. Unfortunately this will only fix cases that use the generic handler.

@lbesnard @bpasquer  There are other handlers that might have the same problem:
./SOOP/SOOP_ASF_SST_XBT/soop_xbt_pipeline/soop_xbt_incoming_handler_dm.sh
./SOOP/SOOP_TMV/incoming_handler.sh
./ABOS/ASFS/incoming_handler.sh
./ABOS/common/incoming_handler.sh
./ANFOG/DM/incoming_handler.sh

Would it be easier to update the `trigger_checkers_and_add_signature` function so that it *always* returns a temp copy of the file?

julian1:
Hi @mhidas,

> Would it be easier to update the trigger_checkers_and_add_signature function so that it always returns a temp copy of the file?

I looked at doing that. The relevant bit of code that could be changed is here,

https://github.com/aodn/data-services/blob/master/lib/common/netcdf-check.sh#L167

I decided against it, since I felt there was a higher chance of other unintended side-effects (bugs)- if the filename was changed for other significant parts of the system.  

Imho all the double file handling is pretty awful, with lots of hidden assumptions and poor cleanup (versus using exception handling etc). It's also difficult to debug without a debugger and when the code keeps crosses language boundaries - between bash and python etc.

If you were confident that making the change to trigger_checkers_and_add_signature() was low-risk and were prepared to do testing, then we could do it that way. Otherwise I'm happy to chase through the other handlers and make corrections if needed.

julian1:
A list of other non-generic handlers that call trigger_checkers_and_add_signature()

```
meteo@debian:~/imos/data-services$ grep -o trigger_checkers_and_add_signature  $( find | grep incoming_handler.sh | grep -v generic )
./ANFOG/DM/incoming_handler.sh:trigger_checkers_and_add_signature
./SOOP/SOOP_TMV/incoming_handler.sh:trigger_checkers_and_add_signature
./ANMN/common/incoming_handler.sh:trigger_checkers_and_add_signature
./ANMN/AM/incoming_handler.sh:trigger_checkers_and_add_signature
./ABOS/ASFS/incoming_handler.sh:trigger_checkers_and_add_signature
./ABOS/common/incoming_handler.sh:trigger_checkers_and_add_signature
```

julian1:
On @mhidas advice, I'll make the change to trigger_checkers_and_add_signature, and deploy to 14-nsp where for more testing.

julian1:
https://github.com/aodn/data-services/compare/fix_410_2

julian1:
@lbesnard @bpasquer @mhidas 

I manually deployed an update to 14-nsp-mel that should make the fix work for all incoming_handlers. 

It would be good if you guys could verify that it's working with your datasets.

mhidas:
Seems to work ok for my handlers.

Note that apart from checking that the pipelines work under normal conditions, you should really test the case when
* `trigger_checkers_and_add_signature` is called with no checks selected, AND
* the subsequent code (e.g. determining dest_path) fails and results in a call to `file_error`

PR #423, opened 2016-04-20
add VRDU8 Panama ship callsign

lbesnard:


lib/python/ship_callsign
```diff
@@ -38,6 +38,7 @@ VNCF      = Cape-Ferguson
 VNSZ      = Spirit-of-Tasmania-2
 VNVR      = Iron-Yandi
 VRCF6     = Santos-Express
+VRDU8     = OOCL-Panama
```
lbesnard: @atkinsn could you please confirm that the VRDU8 ship callsign will be named 'OOCL-Panama' in the vocab without the hyphen ?

atkinsn: Yes @lbesnard, as just discussed :)

PR #424, opened 2016-04-20
Fix 410 2

julian1:
Currently manually deployed on 14-nsp-mel. 

Wait for @lbesnard @bpasquer @mhidas to verify before merging.

See discussion, https://github.com/aodn/data-services/pull/422

### DIFF
--- a/lib/common/netcdf-check.sh
+++ b/lib/common/netcdf-check.sh
@@ -153,8 +153,8 @@ trigger_checkers() {
 }
 export -f trigger_checkers
 
-# trigger netcdf checker for file. if all checks pass, make a temp
-# copy of the file and add checker signature. print temp filename
+# make a writable temp copy of the file and return its full path, and add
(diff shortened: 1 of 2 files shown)


lib/common/netcdf-check.sh
```diff
@@ -165,8 +165,8 @@ trigger_checkers_and_add_signature() {
     trigger_checkers $file $backup_recipient $@
 
     if [ ${#@} == 0 ]; then
-        # no compliance checks triggered, so no signature
-        echo $file
+        # no compliance checks triggered
```
mhidas: Would be good to update the function description above to say that it will always make a writable temp copy of the file and return its full path, and add checker signature if any checks were requested and passed.

mhidas:
:+1: Verified for my part.

lib/common/netcdf-check.sh
julian1: @mhidas done!

PR #427, opened 2016-04-21
ship callsign using poolparty vocab

lbesnard:
* working version to replace previous function reading a text file of
   ship callsign and vessel name equivalence
 * download vocab file from content.aodn.org.au
 * cannot merge yet as vocab is not ready. Waiting for version 1.1 to
   have all vessel names is the same XML formatting
 * If merged now unittests for FHZI vessel will fail wich is a good
   safety net

wip


lbesnard:
FOR review only. 
NOT TO MERGE, waiting for version 1.1 of platform vocab from @atkinsn 

FYI @smancini @bpasquer

lib/python/ship_callsign.py
```diff
@@ -1,42 +1,70 @@
...
+    next_item_exist    = True
+    platform           = {}
+    item               = 0
 
-def _call_parser(conf_file):
-    parser             = SafeConfigParser()
-    parser.optionxform = str  # to preserve case
-    parser.read(conf_file)
-    return parser
+    while next_item_exist:
```
mhidas: This probably works but there's a much more straight-forward way to iterate over a list in Python:
```python
for item in root:
    if 'Description' in item.tag:
        # etc...
```
or 
```python
for item in root.iter():
```
depending on whether you just want the direct children of root or everything below it (https://docs.python.org/2/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.iter)

Then you don't need `next_item_exist` or the try/except at the end of the loop.

lib/python/ship_callsign.py
```diff
@@ -1,42 +1,70 @@
...
+        # main element name we're interested in
+        if 'Description' in root[item].tag:
+            # for every element, iterate over the sub elements and look for
+            # common platform label
+            altLabel_idx  = None
+            prefLabel_idx = None
 
+            for x in range(len(list(root[item]))):
```
mhidas: Same here
```python
for x in item:
```
In fact you might be able to do this in a single loop using the `root.iter()` method above. Read the docs!

lib/python/ship_callsign.py
lbesnard: thx, will give it a go !

lbesnard:
@mhidas thanks for reviewing. followed your suggestions

lib/python/platform_code_vocab.py
```diff
@@ -0,0 +1,49 @@
...
+
+    for item in root:
+        # main element name we're interested in
+        if 'Description' in item.tag:
+            # for every element, iterate over the sub elements and look for
+            # common platform label
+            platform_code = None
+            platform_name = None
+
+            for x, val in enumerate(item):
```
mhidas: Not sure why you need to enumerate. You're not using `x` anywhere.

lib/python/ship_callsign.py
```diff
@@ -4,39 +4,33 @@
...
-    ship_callsign_file = os.path.join(function_dir, 'ship_callsign')
-
-    parser             = _call_parser(ship_callsign_file)
-    ship_callsign      = dict(parser.items('ship_callsign'))
+    platform_codes = platform_code_list()
+    platform_codes = {key: item.replace(' ', '-') for key, item in platform_codes.items()}
 
-    return ship_callsign
+    if 'FHZI' in platform_codes.keys():
+        platform_codes['FHZI'] = 'Astrolabe'
```
mhidas: I assume this is a temporary fix until the vocab is updated?

lbesnard: No ... it's not a temporary fix. see @smancini and @atkinsn for this. But basically, the proper vessel name for FHZI is "L'Astrolabe" and we don't want any "L'" in the folder name

atkinsn:
New version, 1.1 has just been published @lbesnard

lbesnard:
can be merge once all unittests pass

atkinsn:
Version 1.2 is up Mr @lbesnard

lbesnard:
good to go. unittests work. 
@mhidas feel free to have a last look at it . but this can be merged

lib/python/platform_code_vocab.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Find the platform_code platform_name equivalence from poolparty platform vocab
+xml file stored in content.aodn.org.au
+
+How to use:
+    import os, sys
+    sys.path.insert(0, os.path.join(os.environ.get('DATA_SERVICES_DIR'), 'lib'))
+    from python.platform_code_vocab import *
```
mhidas: Just pointing it out again, I don't think this is a good way to set this up. Just because you can do it doesn't mean you should. The idea of environment variables is that you can change them and then run the same code in a different environment. You can't do that if some of your environment is hard-coded in.

All you need to do is `export PYTHONPATH=$DATA_SERVICES_DIR/lib/python`. This is already done for the `projectofficer` user in prod. In the PO box you could just add it to the `env` file, or at the top of your `incoming_handler.sh` script. Then you can just say
```python
from platform_code_vocab import *
```

lib/python/ship_callsign.py
```diff
@@ -1,42 +1,36 @@
...
 """
-Reads the ship_callsign text file in the same directory and returns a
+Reads the ship_callsign from platform vocab and returns
 dictionnary of callsign and vessel names.
 Used for SOOP
-
 How to use:
     import os, sys
     sys.path.insert(0, os.path.join(os.environ.get('DATA_SERVICES_DIR'), 'lib'))
     from python.ship_callsign import ship_callsign_list
```
mhidas: Same here.

lib/python/ship_callsign.py
mhidas: Ok, though you could do this in a more generic way. e.g. replace "L'" with "" in all names, or why not just replace the apostrophe with '-' or '_' like you did for the spaces?

lib/python/ship_callsign.py
```diff
@@ -1,42 +1,36 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 """
-Reads the ship_callsign text file in the same directory and returns a
+Reads the ship_callsign from platform vocab and returns
```
mhidas: You're returning all codes & names from the platform vocab, not just ships, so why not call these functions `platform_name_list` and `platform_name`, and just put them in the `platform_name_vocab` module?

Otherwise, if you do want to have just the ship codes &  names, you need to filter the vocab list.

lib/python/ship_callsign.py
lbesnard: It could be done, but I prefer not to. It's a fix for only the Astrolabe, and no other platform. This is way it is more specific, visual and less room for errors

lib/python/ship_callsign.py
lbesnard: I totally agree with you, BUT :
* it is not possible to filter by vessel. There is no way AFAIK to know if a platform is a vessel or mooring easily
* the only reason why there is this ship_callsign.py file is because the previous version of this file is still used by many data-services SOOP script which look for specific callsign. So in order not to change the different scripts, the ship_callsign.py stayed here. Otherwise, it means fixing everything in this PR which could also be done

lbesnard:
@mhidas followed your suggestions, and used the platform category vocab already created by @atkinsn

lib/python/platform_code_vocab.py
```diff
@@ -0,0 +1,125 @@
...
+                # handle more than 1 url match per category of platform
+                if platform_element_sublabels is not None:
+                    if 'narrowMatch' in platform_element_sublabels:
+                        val_cat_url = val.attrib[val.attrib.keys()[0]]
+                        platform_cat_url_list.append(val_cat_url)
+
+                    elif 'prefLabel' in platform_element_sublabels:
+                        platform_cat = val.text
+
+            if platform_cat is not None and not(not platform_cat_url_list):
```
mhidas: Why the double negative? Should be
```python
if platform_cat is not None and platform_cat_url_list:
```

lib/python/platform_code_vocab.py
```diff
@@ -0,0 +1,125 @@
...
+            platform_cat_url_list = []
+            platform_cat          = None
+
+            for val in item:
+                platform_element_sublabels = val.tag
+
+                # handle more than 1 url match per category of platform
+                if platform_element_sublabels is not None:
+                    if 'narrowMatch' in platform_element_sublabels:
+                        val_cat_url = val.attrib[val.attrib.keys()[0]]
```
mhidas: A simpler way would be `val_cat_url = val.attrib.values()[0]`. Also you have to be careful addressing `dict` objects by index because they are not ordered! If `val.attrib` has more than one item, this will give you a random one. Is that what you want?

lib/python/platform_code_vocab.py
```diff
@@ -0,0 +1,125 @@
...
+                    elif 'prefLabel' in platform_element_sublabels:
+                        platform_cat = val.text
+
+            if platform_cat is not None and not(not platform_cat_url_list):
+                platform_cat_list[platform_cat] = platform_cat_url_list
+
+    response.close()
+    return platform_cat_list
+
+def platform_altlabels_per_preflabel(*args):
```
mhidas: If you're only expecting at most 1 argument, you should ideally use something like `(category = Null)` instead of `(*args)`. That way calling the function with more than 1 arg will raise an exception rather than silently ignoring it.

lib/python/platform_code_vocab.py
```diff
@@ -0,0 +1,125 @@
...
+                # handle more than 1 alternative label for same pref label
+                if platform_element_sublabels is not None:
+                    if 'altLabel' in platform_element_sublabels:
+                        platform_code.append(val.text)
+
+                    elif 'prefLabel' in platform_element_sublabels:
+                        platform_name = val.text
+
+                    elif 'broader' in platform_element_sublabels:
+                        val_cat_url      = val.attrib[val.attrib.keys()[0]]
```
mhidas: `val.attrib.values()[0]`

lib/python/platform_code_vocab.py
```diff
@@ -0,0 +1,125 @@
...
+                        platform_url_cat = val_cat_url
+
+            # use the optional argument
+            if filter_cat_type:
+                if platform_url_cat in filter_cat_url_list:
+                    if platform_name is not None and not(not platform_code):
```
mhidas: double negative again

lbesnard:
thx @mhidas

PR #431, opened 2016-04-26
Fix mime handling on s3 uploads - aodn/portal #2174

julian1:


lib/common/s3.sh
```diff
@@ -68,12 +68,32 @@ s3_put_no_index_keep_file() {
...
     local object_name=$1; shift
     local dst=$S3_BUCKET/$object_name
+    local mime=''
 
     test -f $src || file_error "Not a regular file"
 
     log_info "Moving '$src' -> '$dst'"
 
-    $S3CMD --no-preserve sync $src $dst || file_error "Could not push to S3 '$src' -> '$dst'"
+    if [[ $file =~ \.gz$ ]]; then
```
mhidas: Do you mean `$src` instead of `$file`?

lib/common/s3.sh
```diff
@@ -68,12 +68,32 @@ s3_put_no_index_keep_file() {
...
+        mime='image/png'
+    elif [[ $file =~ \.jpg$ ]]; then
+        mime='image/jpeg'
+    fi
+
+    if [ -z $mime ]; then
+      # use guess mime
+      $S3CMD --no-preserve sync $src $dst || file_error "Could not push to S3 '$src' -> '$dst'"
+    else
+      $S3CMD -m $mime --no-preserve sync $src $dst || file_error "Could not push to S3 '$src' -> '$dst'"
```
pmbohm: @julian1 Excellent fix but could we have a complete set of mime types defined? This will be important for the public S3 bucket
.gif .m .jpeg .mov .doc .xls .zip etc etc

julian1:
See other discussion here, https://github.com/aodn/aodn-portal/issues/2174

Chef already configures the s3cmd client with the following settings. 

```
default_mime_type = binary/octet-stream
guess_mime_type = True
```


It should therefore work to correctly determine mime types - barring bugs in the s3client and python-magic, that are noted.

My fix just forces the obvious cases and for the cases that we know are not working - csv and I think pdf.

Probably the proper way is to fix s3cmd so that it reads /etc/mime.types, but I don't think we have the resources to do that properly for the moment.

lib/common/s3.sh
julian1: Uggh. fixed.

pmbohm:
@julian1 Well I'm 100% sure that the .m extension is not working as we want it
Even if  `guess_mime_type = True` works (probably not) , S3 is using `http://www.iana.org/assignments/media-types/media-types.xhtml` and people have commented that .jpg is missing for example!!!!
`http://docs.aws.amazon.com/AWSImportExport/latest/DG/FileExtensiontoMimeTypes.html`

julian1:
@pmbohm Can you give me a link to an uploaded .m file.

I confess I don't know why such a file would be being pushed through the pipeline.

Could any project-officers please comment on expected file types.

pmbohm:
I think I will leave this to project officers to merge. @mhidas @lbesnard @xhoenner @bpasquer @smancini 
This is an issue that may not go away quickly requiring a proper fix by Amazon, therefore project officers will need to be aware of. 
Existing files may will need to be fixed using a S3 browser or some such

pmbohm:
@julian1 It in the original bug as a comment 
http://imos-data.s3-website-ap-southeast-2.amazonaws.com/IMOS/ANMN/Acoustic/code/NL_load_logger_data_new.m
@mhidas

julian1:
I would like to suggest this PR is merged as-is, since it directly addresses portal bug https://github.com/aodn/aodn-portal/issues/2174.

If other file-types are discovered that need special case-handling, then either,

- Project-officers can make the change as needed (I think it's clear how the code works).
- JPSR can decide to make it a planned task.

I've spent a couple of days on this already. If I am to spend more time, then I feel that the work should be done in a context other than a bug-fix.

pmbohm:
Existing files wont be fixed by this pull request. So aodn/aodn-portal#2174 remains open. 
I have created a backlog item to address a larger solution https://github.com/aodn/backlog/issues/325

julian1:
@pblain @smancini 

This PR addresses the issue for any new files uploaded to s3. 

For current S3 objects that have incorrect mime types, it will be necessary to either reingest them in pipeline, or write a script to iterate all S3 objects and update.

julian1:
> So aodn/aodn-portal#2174 remains open. 

There is nothing standing in the way of project-officers reingesting files or alternatively writing a script to fix existing files.

If there is time at the end of this iteration and JPSR agree, then developers could also write the script.

lib/common/s3.sh
```diff
@@ -68,12 +68,32 @@ s3_put_no_index_keep_file() {
...
+    elif [[ $src =~ \.jpg$ ]]; then
+        mime='image/jpeg'
+    fi
+
     log_info "Moving '$src' -> '$dst'"
 
-    $S3CMD --no-preserve sync $src $dst || file_error "Could not push to S3 '$src' -> '$dst'"
+    if [ -z $mime ]; then
+      # use guess mime
+      $S3CMD --no-preserve sync $src $dst || file_error "Could not push to S3 '$src' -> '$dst'"
```
lbesnard: indentation

lib/common/s3.sh
```diff
@@ -68,12 +68,32 @@ s3_put_no_index_keep_file() {
...
+        mime='application/octet-stream'
+    elif [[ $src =~ \.csv$ ]]; then
+        mime='text/csv'
+    elif [[ $src =~ \.pdf$ ]]; then
+        mime='application/pdf'
+    elif [[ $src =~ \.png$ ]]; then
+        mime='image/png'
+    elif [[ $src =~ \.jpg$ ]]; then
+        mime='image/jpeg'
+    fi
```
lbesnard: had a chat with @pmbohm , but maybe we could have a special octet-stream for any other extension. Anyway, this seems really annoying, and having to reprocess all files because of this will take some time

julian1: @lbesnard @pblain 

> Maybe we could have a special octet-stream for any other extension.

The default is already octet-stream for any other extension.

> and having to reprocess all files because of this will take some time


I have java code that can update metadata recursively across all s3 keys. 

If this had been merged as I suggested then I could have updated everything at the end of last iteration when I had nothing else to do.

Issue is closed now, so it's your problem.

lbesnard: > Issue is closed now, so it's your problem.

thanks but this is not really constructive. I think no one from the PO team merged this because we just don't know anything re Apache. We have no knowledge on this regard, can't test it and rely on the dev team to tell us if this is good to go or not. So it's up to another Dev to merge this in my opinion.

pmbohm:
Merging this as a band-aid fix for at least a few file types. I'll be amazed if we don't see a very similar bug report re-appear in the future.

PR #433, opened 2016-04-27
s3 functions - list recursively objects and download

lbesnard:
- created 2 functions to ease the process of folder restructuration,
  download ...
- file objects can be listed recursively with s3_ls_recur function
- file objects can be downloaded recursively with
  s3_dl_recur_latest_version function

### DIFF
--- a/lib/common/s3.sh
+++ b/lib/common/s3.sh
@@ -94,6 +94,7 @@ export -f s3_ls
 # $1 - path of folder on s3
 s3_ls_recur() {
     local folder=${1%/}; shift # remove trailing slash
+    [ -z "$folder" ] && { echo "Folder object not specified"; return 1; }
     local object
 
     for object in `s3lsv -a -b imos-data -k $folder/`; do
@@ -114,7 +115,7 @@ export -f s3_ls_recur
 s3_dl_recur_latest_version() {
     local folder=$1; shift
     local output_folder=$1; shift
-    [ -z "$output_folder" ] && { echo "output_folder is empty"; return 1; }
+    [ -z "$output_folder" ] && { echo "No output_folder specified"; return 1; }
     mkdir -p $output_folder  || return 1
 
     # only keeping uniq filename objects
@@ -131,7 +132,6 @@ s3_dl_recur_latest_version() {
 
         obj_vers=`echo "$obj_info" | awk {'print $1'}`
         obj_name=`echo "$obj_info" | awk {'print $3'}`
-        obj_date=`echo "$obj_info" | awk {'print $2'}`
 
         mkdir -p `dirname $output_folder/$obj_name`
         echo "Downloading $obj_name"
--- a/lib/common/s3.sh
+++ b/lib/common/s3.sh
@@ -89,3 +89,53 @@ s3_ls() {
     $S3CMD ls $S3_BUCKET/$path/ | tr -s " " | cut -d' ' -f4 | sed -e "s#^$S3_BUCKET/$path/##"
 }
 export -f s3_ls
+
+# list recursively a folder on s3 and returns list of fullpath of file objects
+# $1 - path of folder on s3
+s3_ls_recur() {
+    local folder=${1%/}; shift # remove trailing slash
+    local object
+
+    for object in `s3lsv -a -b imos-data -k $folder/`; do
+        # check if object is a folder
+        if `echo "$object" | grep -E -q '\/$'`; then
+            s3_ls_recur "$object"
+        else
+            echo "$object" | grep -o -E "$folder.*$"
+        fi
+    done
+}
+export -f s3_ls_recur
+
+# download recursively s3 objects from a given folder object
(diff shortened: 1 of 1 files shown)


lbesnard:
@pblain This is the kind of PR which would worth a look by the dev's team
@smancini  FYI

lib/common/s3.sh
```diff
@@ -89,3 +89,53 @@ s3_ls() {
     $S3CMD ls $S3_BUCKET/$path/ | tr -s " " | cut -d' ' -f4 | sed -e "s#^$S3_BUCKET/$path/##"
 }
 export -f s3_ls
+
+# list recursively a folder on s3 and returns list of fullpath of file objects
+# $1 - path of folder on s3
+s3_ls_recur() {
+    local folder=${1%/}; shift # remove trailing slash
```
mhidas: Should also check that `$folder` is not empty! I tried running this with no arguments and got strange behaviour (it listed a bunch of SOOP-BA paths in `/mnt/imos-data/`.

lib/common/s3.sh
```diff
@@ -89,3 +89,53 @@ s3_ls() {
...
+    local object
+
+    for object in `s3lsv -a -b imos-data -k $folder/`; do
+        # check if object is a folder
+        if `echo "$object" | grep -E -q '\/$'`; then
+            s3_ls_recur "$object"
+        else
+            echo "$object" | grep -o -E "$folder.*$"
```
mhidas: Don't you want to filter out just the current versions? E.g. compare the output of 
`s3_ls_recur IMOS/ANMN/NRS/REAL_TIME/NRSMAI/Wave` (195 lines) to 
`ls /mnt/imos-data/IMOS/ANMN/NRS/REAL_TIME/NRSMAI/Wave` which shows there are only 3 files there.

lib/common/s3.sh
```diff
@@ -89,3 +89,53 @@ s3_ls() {
...
+export -f s3_ls_recur
+
+# download recursively s3 objects from a given folder object
+# ex: s3_dl_recur_latest_version UWA /tmp/download_test
+# $1 - path of folder on s3
+# $2 - path of output folder where files will be downloaded keeping same structure
+s3_dl_recur_latest_version() {
+    local folder=$1; shift
+    local output_folder=$1; shift
+    [ -z "$output_folder" ] && { echo "output_folder is empty"; return 1; }
```
mhidas: "No ouptut folder specified" would be a more accurate error message.

lib/common/s3.sh
```diff
@@ -89,3 +89,53 @@ s3_ls() {
...
+    local file_list=`s3_ls_recur $folder | uniq | sort`
+    local object
+    local obj_info
+
+    for object in $file_list; do
+        obj_info=`s3lsv -a -b imos-data -k $object`
+        # if more than one version of same object, we reverse sort the second column(date) and take the first line
+        # i.e. the lastest object version
+        # date format is yyyy-mm-ddTHH:MM:SS
+        obj_info=`echo $obj_info | sort -b -r -k 2.1,2.4 -k 2.6,2.7 -k 2.9,2.10 -k 2.12,2.13 -k 2.15,2.16 -k 2.18,2.19 | head -1`
```
mhidas: No need to sort. `s3lsv` output is already sorted. Current version is the first line.

lib/common/s3.sh
```diff
@@ -89,3 +89,53 @@ s3_ls() {
...
+
+        obj_vers=`echo "$obj_info" | awk {'print $1'}`
+        obj_name=`echo "$obj_info" | awk {'print $3'}`
+        obj_date=`echo "$obj_info" | awk {'print $2'}`
```
mhidas: Don't need `obj_date`

lib/common/s3.sh
lbesnard: I think it would be more up to the s3lsv python function https://github.com/aodn/chef/blob/master/cookbooks/imos_po/files/default/s3lsv to handle this with a different optional argument to only list objects which exists(visible). Doing this here would be patchy and couldn't be re-use in my opinion while it would actually be great to have it inside s3lsv

lib/common/s3.sh
lbesnard: hum i look again because it didn't seem like it to me

lib/common/s3.sh
lbesnard: but actually can be done fine. yeap I'll add this

lbesnard: @mhidas I wasn't aware the bucket was 'mounted' and accessible via $DATA_DIR. I'm now wondering if these functions are necessary.

mhidas: It was meant to be a temporary thing, though now that Dan's gone it will probably be around for a while.

lbesnard: ok. So if it's a temporary thing, it would not be really smart to check a file exists in this dir

bhasin85:
@lbesnard @mhidas Can we close/merge this PR?

PR #434, opened 2016-04-27
Turn off CF checks for ANMN_NSW

mhidas:
I need to temporarily turn this off to publish some files in the incoming backlog that don't pass the checkers.


mhidas:
@lbesnard @bpasquer Could you please merge this? It won't take more than 30sec!

mhidas:
@bpasquer Thanks!

PR #436, opened 2016-04-28
Turn CF checks back on for ANMN_NSW

mhidas:
Reverts aodn/data-services#434


mhidas:
Backlog of old files has been cleared. New files should pass CF. (Not yet IMOS because we need IMOS-1.4 checker.)

PR #437, opened 2016-04-28
SRS OC - handle new file naming convention

lbesnard:


lbesnard:
L2OC_BASE. string is removed

PR #445, opened 2016-05-04
Fix typo on global attribute acknowledgement. NOBS1/2 cannot be ancil…

ggalibert:
…lary_variables without using a standard_name modifier which I will not try to do at this stage. NOBS1/2 variable attribute units needs to be set to 1.


bpasquer:
Good to go @ggalibert 👍

PR #447, opened 2016-05-04
QC variable attribute quality_control_conventions needed fixing. Glob…

ggalibert:
…al attribute quality_control_set can be removed. Global attribute distribution_statement needed fixing. Global attribute comment can be updated.

### DIFF
--- a/ACORN/current_generator/acorn_constants.py
+++ b/ACORN/current_generator/acorn_constants.py
@@ -384,7 +384,7 @@ current_variables = {
             [ "standard_name",               "eastward_sea_water_velocity status_flag" ],
             [ "long_name",                   "quality flag for eastward_sea_water_velocity" ],
             [ "coordinates",                 "TIME LATITUDE LONGITUDE" ],
-            [ "quality_control_conventions", "IMOS standard set using IODE flags" ],
+            [ "quality_control_conventions", "IMOS standard set using the IODE flags" ],
             [ "quality_control_set",         1. ],
             [ "valid_min",                   min(qc_flag_values) ],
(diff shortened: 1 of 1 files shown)


ACORN/current_generator/acorn_constants.py
```diff
@@ -450,8 +449,8 @@
...
     [ "acknowledgement",              "Data was sourced from the Integrated Marine Observing System (IMOS) - IMOS is supported by the Australian Government through the National Collaborative Research Infrastructure Strategy (NCRIS) and the Super Science Initiative (SSI)." ],
-    [ "distribution_statement",       "Data, products and services from IMOS are provided \"as is\" without any warranty as to fitness for a particular purpose" ],
-    [ "comment",                      "${extraComment}This NetCDF file has been created using the IMOS NetCDF User Manual v1.2. A copy of the document is available at http://imos.org.au/facility_manuals.html ." ]
+    [ "distribution_statement",       "Data may be re-used, provided that related metadata explaining the data has been reviewed by the user, and the data is appropriately acknowledged. Data, products and services from IMOS are provided \"as is\" without any warranty as to fitness for a particular purpose." ],
+    [ "comment",                      "${extraComment}This NetCDF file has been created using the IMOS NetCDF Conventions v1.4." ]
```
mhidas: This should be reflected in the `Conventions` attribute. It looks like you're still setting that to "CF-1.5,IMOS-1.2". In fact, if you update that, you don't really need this comment.

ggalibert: Well spotted, thanks. I will update Conventions but I think I'll keep the comment as is to have a reference of the name of the document.

PR #453, opened 2016-05-09
AUV processing

lbesnard:
* fix reporting function
 * cleaner EPSG convertion


lbesnard:
@mhidas followed your suggestion

PR #454, opened 2016-05-09
LFTP - get addition fix

lbesnard:
* the get_lftp_additions looekd for a very specific type of string
   from lftp logs'get -O ...'. This wouldn' let other arguments be such
   as 'get -e -O'. For this reason, None of AATAMS DM updated files got
   pushed to the incoming dir
 * pass unittests

### DIFF
--- a/lib/common/sync.sh
+++ b/lib/common/sync.sh
@@ -34,7 +34,11 @@ get_lftp_additions() {
     local lftp_output_file=$1; shift
     local local_base=$1; shift
 
-    grep "^get -O " $lftp_output_file | _parse_lftp_file_addition_line | \
+    # see definition http://lftp.yar.ru/lftp-man.html
+    # sample line can be :
+    # get -e -O /tmp/argo/dac/csio/2900322 ftp://ftp.ifremer.fr/ifremer/argo/dac/csio/2900322/2900322_Rtraj.nc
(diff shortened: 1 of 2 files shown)


lbesnard:
@xhoenner

lbesnard:
still open for review

lbesnard:
@xhoenner well it passes unittests. I'll just merge it I guess

lib/common/sync.sh
```diff
@@ -56,13 +60,15 @@ export -f get_lftp_deletions
...
         line=(${line// / })
         # sample line is:
         # get -O /tmp/argo/dac/csio/2900322 ftp://ftp.ifremer.fr/ifremer/argo/dac/csio/2900322/2900322_Rtraj.nc
+        # get -e -O /tmp/argo/dac/csio/2900322 ftp://ftp.ifremer.fr/ifremer/argo/dac/csio/2900322/2900322_Rtraj.nc
         # we take the basename from the `ftp://` part and add it to the
         # /tmp/argo/dac/... part
-        local path_basename=`basename ${line[3]}`
-        echo ${line[2]}/$path_basename
+        local path_basename=`basename ${line[-1]}`
+        echo $path_dir/$path_basename
```
jonescc: Why not use ${line[-2]} as per the previous line - using 2 methods of cutting up the source line here which means more work deciphering what you are doing  - I'd use one or the other not both

PR #461, opened 2016-05-17
Remove RAMADDA from repo

lbesnard:


lbesnard:
@pblain This needs to go. Could you confirm please ?

pblain:
@lbesnard - it can go.  MARVL no longer uses it.

PR #462, opened 2016-05-18
Burst average python conversion

lbesnard:


lbesnard:
NOT TO MERGE : WIP

First WQM file generated with Python script
[IMOS_ANMN-NRS_KOSTUZ_20130806T060001Z_NRSNIN_FV02_NRSNIN-1307-WQM-46-burst-averaged_END-20140213T084502Z_C-20160518T200838Z.nc.tar.gz](https://github.com/aodn/data-services/files/270181/IMOS_ANMN-NRS_KOSTUZ_20130806T060001Z_NRSNIN_FV02_NRSNIN-1307-WQM-46-burst-averaged_END-20140213T084502Z_C-20160518T200838Z.nc.tar.gz)

FV01 WQM source is http://thredds.aodn.org.au/thredds/dodsC/IMOS/ANMN/NRS/NRSNIN/Biogeochem_timeseries/IMOS_ANMN-NRS_KOSTUZ_20130806T060000Z_NRSNIN_FV01_NRSNIN-1307-WQM-46_END-20140212T070000Z_C-20140430T022104Z.nc.html

FV02 Burst product done previously by Monique :
http://thredds.aodn.org.au/thredds/catalog/IMOS/ANMN/NRS/NRSNIN/Biogeochem_timeseries/burst-averaged/catalog.html?dataset=IMOS/ANMN/NRS/NRSNIN/Biogeochem_timeseries/burst-averaged/IMOS_ANMN-NRS_KOSTUZ_20130806T000000Z_NRSNIN_FV02_NRSNIN-1307-WQM-46-burst-averaged_END-20140211T235900Z_C-20150318T081000Z.nc

Open to suggestions/review ... Still WIP

lbesnard:
@mhidas @ggalibert @smancini 

I've managed to create WQM files quite close to Monique's old ones. However I've noticed some difference in the values we both exclude. Especially towards the end of every variable :

Looking at existing Burst average files for NRSNIN:
http://thredds.aodn.org.au/thredds/catalog/IMOS/ANMN/NRS/NRSNIN/Biogeochem_timeseries/burst-averaged/catalog.html

The end date in the filename is always before midnight which is a bit suspicious ```END-20120223T235900Z_C-20140605T092700Z.nc```

while on the other end, I keep all data before the  ```time_deployment_end``` attribute. So I guess there is a mistake in Monique's code, or maybe there was a good reason for it.

This means I always create more data points than Monique's code does.

ggalibert:
There is indeed discrepancy between time_coverage_start/end or time_deployment_start/end and the first/last time stamp in Monique's files. It would be interesting to find out where does this error stem from in her original code.

ANMN/burst_averaged_product/burstproduct/burst_average_product.py
```diff
@@ -0,0 +1,382 @@
...
+
+- for CTD file
+http://thredds.aodn.org.au/thredds/catalog/IMOS/ANMN/NRS/NRSKAI/CTD_timeseries/catalog.html
+- remove VOLT, TIMESERIES, NOMINAL_DEPTH????
+- add all variables to existing conf file. Should work also for WQM file
+- refactore 1 or 2 WQM function as CTD will be very similar
+"""
+
+
+def get_input_file_rel_path(input_netcdf_file_path):
```
mhidas: Seems unnecessarily complicated to use a shell subprocess to call another Python script! The guts or `ANMN/common/dest_path.py` is actually a class defined in `lib/python/file_classifier.py`, so you can just do this:
```python
from file_classifier import MooringFileClassifier

file_path = MooringFileClassifier.dest_path(input_netcdf_file_path)
```

ANMN/burst_averaged_product/burstproduct/burst_average_product.py
```diff
@@ -0,0 +1,382 @@
...
+            return ''
+            # sys.exit(1)
+    else:
+        return ''
+
+def get_burst_duration(netcdf_file_path):
+    """
+    returns the burst duration as float defined in the NetCDF file if exists
+    """
+    netcdf_file_obj = Dataset(netcdf_file_path, 'r', format='NETCDF4')
```
mhidas: Don't need to specify format if you're opening an existing file for reading.

ANMN/burst_averaged_product/burstproduct/burst_average_product.py
lbesnard: thx Marty !

ANMN/burst_averaged_product/burstproduct/burst_average_product.py
mhidas: You could save repeatedly opening and closing the file by just opening it in the main program and passing the `Dataset` object around instead of the filename.

ANMN/burst_averaged_product/burstproduct/burst_average_product.py
```diff
@@ -0,0 +1,382 @@
...
+    if hasattr(netcdf_file_obj, 'instrument_burst_duration'):
+        burst_duration = float(netcdf_file_obj.instrument_burst_duration)
+    else:
+        burst_duration = 60.0
```
mhidas: There's a convenient shorthand for the above 4 lines:
```python
burst_duration = getattr(netcdf_file_obj, 'instrument_burst_duration', 60.0)
```
The third argument is a default, returned if the attribute doesn't exist.

ANMN/burst_averaged_product/burstproduct/burst_average_product.py
```diff
@@ -0,0 +1,401 @@
...
+    time_before = date_deployment_start >= time[:]
+    time_after  = date_deployment_end <= time[:]
+
+    var_qc_flag_exclusion = [a or b for a, b in zip(time_before, var_qc_flag_exclusion)]
+    var_qc_flag_exclusion = [a or b for a, b in zip(time_after, var_qc_flag_exclusion)]
+
+    netcdf_file_obj.close()
+    return var_values, var_qc_flag_exclusion
+
+def list_var_not_to_average(netcdf_file_path):
```
mhidas: Shouldn't it be called `list_var_to_average`?

lbesnard: :+1:

ANMN/burst_averaged_product/burstproduct/burst_average_product.py
```diff
@@ -0,0 +1,401 @@
...
+    """
+    create burst data from all vars available in netcdf
+    """
+    time_values              = get_time_val(netcdf_file_path)
+    varname_to_burst_average = list_var_not_to_average(netcdf_file_path)
+    burst_duration           = get_burst_duration(netcdf_file_path)
+    burst_vars               = {}
+
+    for varname in varname_to_burst_average:
+        [var_values, var_qc_flag_exclusion] = get_var_val_var_qc_exclusion(netcdf_file_path, varname, min_bad_qc_flag(varname))
```
mhidas: You don't need `[]` to unpack a tuple, i.e. `var_values, var_qc_flag_exclusion = ...` does exactly the same thing.

Also, it would be nice to wrap such long lines to something more legible, e.g.
```python
        var_values, var_qc_flag_exclusion = get_var_val_var_qc_exclusion(
            netcdf_file_path, varname, min_bad_qc_flag(varname)
        )
```
(See [style guide](https://www.python.org/dev/peps/pep-0008/#id17))

ANMN/burst_averaged_product/burstproduct/burst_average_product.py
```diff
@@ -0,0 +1,401 @@
...
+        [time_mean_burst, var_mean_burst, var_min_burst, var_max_burst, var_std_burst, var_num_obs] = burst_average_data(time_values, var_values, burst_duration, var_qc_flag_exclusion)
+
+        burst_vars[varname] = [time_mean_burst, var_mean_burst, var_min_burst, var_max_burst, var_std_burst, var_num_obs]
```
mhidas: No point unpacking all these variables if you don't need each element on its own. The above two lines do the same as:
```python 
        burst_vars[varname] = list(
            burst_average_data(time_values, var_values, burst_duration, var_qc_flag_exclusion)
        )
```
... and you may not even need the `list()` if the rest of the code can handle a tuple.

ANMN/burst_averaged_product/burstproduct/burst_average_product.py
```diff
@@ -0,0 +1,401 @@
...
+        var_qc_flag_exclusion = netcdf_file_obj.variables['%s_quality_control' % varname][:] >= qc_flag
+
+    # we exclude as well ALL TIMES before date_deployment_start, and after
+    # date_deployment_end
+    time                  = netcdf_file_obj.variables['TIME']
+    date_deployment_start = date2num(datetime.strptime(netcdf_file_obj.time_deployment_start, '%Y-%m-%dT%H:%M:%SZ'), time.units, time.calendar)
+    date_deployment_end   = date2num(datetime.strptime(netcdf_file_obj.time_deployment_end, '%Y-%m-%dT%H:%M:%SZ'), time.units, time.calendar)
+
+    time_before = date_deployment_start >= time[:]
+    time_after  = date_deployment_end <= time[:]
```
mhidas: Don't need to repeat this for each data variable.

ANMN/burst_averaged_product/burstproduct/burst_average_product.py
```diff
@@ -0,0 +1,401 @@
...
+    for a qc flag values [0:9], returns the var_values array, and var_qc_flag_exclusion
+    which is a boolean array returning True for all qc >= qc_flag
+
+    Also exlude all TIME values outside of deployment range as defined in gatts
+    """
+    netcdf_file_obj       = Dataset(netcdf_file_path, 'r')
+    shape_var_array = netcdf_file_obj.variables[varname].shape
+
+    if len(shape_var_array) == 3:
+        # WQM files
```
mhidas: This is not a differene between WQM and CTD files. It depends on which version of the Toolbox processed it.

lbesnard: thanks. I've actually handled it better in next commit

ANMN/burst_averaged_product/burstproduct/burst_average_product.py
```diff
@@ -0,0 +1,401 @@
...
+    difft         = np.round(np.diff(time_values) * n_seconds_day) # this gives us a spike for each new burst
+    allowed_gap   = burst_duration
+    fill_value    = float('nan')
+
+    # look for start index of each burst, ie the index of each new spike of
+    # difft variable
+    idx_spike = [0] # initialise first index
+    for idx, val_boolean in enumerate(difft < allowed_gap):
+        if not val_boolean:
+            idx_spike.append(idx + 1)
```
mhidas: This is also something you don't need to repeat for each variable, as it only depends on the timestamps.

Also, I find this a bit more straightforward:
```python
spikes = np.where(difft >= allowed_gap)[0] + 1
idx_spike.extend(spikes)
```

lbesnard:
Have to investigate issue with file :
IMOS_ANMN-NRS_ACESTZ_20100827T010136Z_NRSKAI_FV01_NRSKAI-1009-NXIC-CTD-103.5_END-20110316T030200Z_C-20151116T060212Z.nc

Get this error message without any python error, line error
```
HDF5: infinite loop closing library
      R,D,G,A,S,T,F,FD,P,FD,P,FD,P,E,E,SL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL
```

ANMN/burst_averaged_product/burst_average_product.py
```diff
@@ -0,0 +1,388 @@
...
+        gatt = 'comment'
+        setattr(output_netcdf_obj, gatt, '%s. %s' % (getattr(input_netcdf_obj, gatt), getattr(output_netcdf_obj, gatt) ))
+
+    # add values to variables
+    output_netcdf_obj['TIME'][:]      = np.ma.masked_invalid(time_burst_vals)
+    output_netcdf_obj['LATITUDE'][:]  = np.ma.masked_invalid(input_netcdf_obj['LATITUDE'][:])
+    output_netcdf_obj['LONGITUDE'][:] = np.ma.masked_invalid(input_netcdf_obj['LONGITUDE'][:])
+
+    output_netcdf_obj.close()
+    input_netcdf_obj.close()
```
lbesnard: @mhidas 
Could you please test this script on your PO box ? I somehow get a segmentation fault, apparently due closing file objects. And it works well on my host

ANMN/burst_averaged_product/burst_average_product.py
```diff
@@ -0,0 +1,388 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Burst Average Product creation for WQM and CTD FV01 files
+./burst_average_product.py input_netcdf.nc /ouput_dir
```
mhidas: You could keep the script name short and just call it "burst_average.py".  Otherwise it should be "burst_average**d**_product.py".

ANMN/burst_averaged_product/burst_average_product.py
```diff
@@ -0,0 +1,388 @@
...
+def generate_netcdf_burst_filename(netcdf_file_obj, burst_vars):
+    """
+    generate the filename of a burst average netcdf for both CTD and WQM
+    """
+    site_code         = netcdf_file_obj.site_code
+    input_netcdf_name = os.path.basename(input_netcdf_file_path)
+    pattern           = re.compile("^(IMOS_.*)_([0-9]{8}T[0-9]{6}Z)_(.*)_(FV0[0-9])_(.*)_END")
+    match_group       = pattern.match(input_netcdf_name)
+
+    time_burst_vals = burst_vars['TEMP'][0]
```
mhidas: Do you mean `burst_vars['TIME']` here?

ANMN/burst_averaged_product/burst_average_product.py
```diff
@@ -0,0 +1,388 @@
...
+    # remove the first common NaN to all variables
+    for var in burst_vars.keys():
+        for product in range(len(burst_vars[var])):
+            burst_vars[var][product] = burst_vars[var][product][min_index:-1]
+
+    # look for last non NaN. we just have to reverse the list
+    max_non_nan_idx = None
+    for var in burst_vars.keys():
+        var_mean_burst       = burst_vars[var][1] # first non TIME product
+        var_mean_burst       = pd.Series(var_mean_burst) # create a pandas series to find easily last non nan index
```
mhidas: You can get the last non-nan index like this `max( np.where(np.isfinite(var_mean_burst))[0] )`
Otherwise you need to add `pandas` to the requirements.txt file for the repo.

ANMN/burst_averaged_product/burst_average_product.py
```diff
@@ -0,0 +1,388 @@
...
+
+    shutil.move(output_netcdf_file_path, output_dir)
+    shutil.rmtree(tmp_netcdf_dir)
+    return os.path.join(output_dir, os.path.basename(output_netcdf_file_path))
+
+
+if __name__ == "__main__":
+    if len(sys.argv) != 3:
+        msg = 'Usage: %s INPUT_FILE OUTPUT_DIR\n' % sys.argv[0]
+        raise RuntimeError(msg)
```
mhidas: Actually, you don't really need to raise an error here. Just print the message to stderr and exit.

ANMN/burst_averaged_product/generate_ctd_nc_file_gatt
```diff
@@ -0,0 +1,4 @@
+[global_attributes]
+abstract = SAIMOS NRS mooring location. This mooring collects times series data of water currents (x,y,z), CTD (fluorescence, PAR, DO & turbidity). SARDI maintains the SAIMOS mooring network. Data from bursts have been cleaned and averaged to create data products. This file is one such product.
```
mhidas: abstract and keywords should not be hard-coded. You can just append the relevant info (e.g. last two sentences in abstract) to whatever's in the input file.

ANMN/burst_averaged_product/generate_nc_file_att
```diff
@@ -0,0 +1,227 @@
+[global_attributes]
+Conventions             = CF-1.6,IMOS-1.3
+acknowledgement         = Any users of IMOS data are required to clearly acknowledge the source of the material derived from IMOS in the format: "Data was sourced from the Integrated Marine Observing System (IMOS) - IMOS is a national collaborative research infrastructure, supported by the Australian Government." If relevant, also credit other organisations involved in collection of this particular datastream (as listed in "credit" in the metadata record).
+author_email            = laurent.besnard@utas.edu.au
+author_name             = Besnard, Laurent
+citation                = The citation in a list of references is: "IMOS [year-of-data-download], [Title], [data-access-URL], accessed [date-of-access].".
+comment                 = Geospatial vertical min/max information has been computed using the Gibbs-SeaWater toolbox (TEOS-10) v3.02 from latitude and relative pressure measurements (calibration offset usually performed to balance current atmospheric pressure and acute sensor precision at a deployed depth).
```
mhidas: Don't hard-code `comment`. Copy what's in the input file.
Same for most of the other attributes here, especially things that can vary from file to file:
`institution`
`institution_references`
`references`
variable `comment` attributes

ANMN/burst_averaged_product/generate_wqm_nc_file_gatt
```diff
@@ -0,0 +1,4 @@
+[global_attributes]
```
mhidas: Same here. Don't hardcode abstract and keywords. What you add should be the same for WQM and CTD.

ANMN/burst_averaged_product/burst_average_product.py
mhidas: I get a segfault too. Both on my machine and in the PO box.

ANMN/burst_averaged_product/burst_average_product.py
lbesnard: no, because the keys of burst_vars are only variable to average. 
time -> burst_vars['TEMP'][0]
mean_temp -> burst_vars['TEMP'][1]
...

ggalibert: Could you then use more explicit keys than 0, 1, 2, etc...?

time_burst -> burst_vars['TEMP']['time_burst']
mean_temp -> burst_vars['TEMP']['mean']

lbesnard: done. will be visible in next commit

lbesnard:
@mhidas could you please try ? if you do a provision of your box I think the seg fault should disappear. Caused by a bug in the NetCDF4 lib

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -0,0 +1,397 @@
...
+    tmp_netcdf_dir   = tempfile.mkdtemp()
+
+    output_netcdf_file_path = os.path.join(tmp_netcdf_dir, generate_netcdf_burst_filename(input_netcdf_obj, burst_vars))
+    output_netcdf_obj       = Dataset(output_netcdf_file_path, "w", format="NETCDF4")
+
+    # read gatts from input, add them to output. Some gatts will be
+    # overwritten
+    input_gatts     = input_netcdf_obj.__dict__.keys()
+    gatt_to_dispose = ['author', 'toolbox_input_file', 'file_version_quality_control', 'quality_control_set',
+                       'history', 'CoordSysBuilder_', 'netcdf_filename', 'metadata', 'featureType']
```
mhidas: I would keep 'history' and 'featureType' as they are relevant to the output file.

lbesnard: I think that if I added featureType, the files wouldn't pass the CF checker

mhidas: On the other hand you should probably remove 'compliance_checker_version' and 'compliance_checker_last_updated', as these are specific to the original file. They will be added again when the product file passes the checker.

mhidas:
@lbesnard Yep, runs fine now. Good work!

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -0,0 +1,397 @@
...
+    output_netcdf_obj.geospatial_lat_min = input_netcdf_obj['LATITUDE'][:].min()
+    output_netcdf_obj.geospatial_lat_max = input_netcdf_obj['LATITUDE'][:].max()
+    output_netcdf_obj.geospatial_lon_min = input_netcdf_obj['LONGITUDE'][:].min()
+    output_netcdf_obj.geospatial_lon_max = input_netcdf_obj['LONGITUDE'][:].max()
+
+    output_netcdf_obj.geospatial_vertical_min = min(burst_vars['DEPTH']['var_mean'])
+    output_netcdf_obj.geospatial_vertical_max = max(burst_vars['DEPTH']['var_mean'])
+
+    output_netcdf_obj.time_coverage_start = num2date(time_burst_vals, input_netcdf_obj['TIME'].units, input_netcdf_obj['TIME'].calendar).min().strftime('%Y-%m-%dT%H:%M:%SZ')
+    output_netcdf_obj.time_coverage_end   = num2date(time_burst_vals, input_netcdf_obj['TIME'].units, input_netcdf_obj['TIME'].calendar).max().strftime('%Y-%m-%dT%H:%M:%SZ')
```
mhidas: You could make this neater and more readable by doing the `num2date` just once and then using the result in the above 2 lines.

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -0,0 +1,397 @@
...
+
+    # set up dimensions
+    output_netcdf_obj.createDimension("TIME", len(time_burst_vals))
+
+    # set up variables
+    var_time = output_netcdf_obj.createVariable("TIME", "d", ("TIME",), fill_value=get_imos_parameter_info('TIME', '_FillValue'))
+    var_lat  = output_netcdf_obj.createVariable("LATITUDE", "d", fill_value=get_imos_parameter_info('LATITUDE', '_FillValue'))
+    var_lon  = output_netcdf_obj.createVariable("LONGITUDE", "d", fill_value=get_imos_parameter_info('LONGITUDE', '_FillValue'))
```
mhidas: LAT & LON can just be type float.

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -0,0 +1,397 @@
...
+    for var in burst_vars.keys():
+        fillvalue = get_imos_parameter_info(var, '_FillValue')
+        if fillvalue == []:
+            fillvalue = 999999.0
+
+        output_netcdf_obj.createVariable(var, "f4", ("TIME",), fill_value=fillvalue)
+        output_netcdf_obj.createVariable('%s_burst_min' % var, "f4", ("TIME",), fill_value=fillvalue)
+        output_netcdf_obj.createVariable('%s_burst_max' % var, "f4", ("TIME",), fill_value=fillvalue)
+        output_netcdf_obj.createVariable('%s_burst_sd' % var, "f4", ("TIME",), fill_value=fillvalue)
+        output_netcdf_obj.createVariable('%s_num_obs' % var, "i4", ("TIME",), fill_value=fillvalue)
```
mhidas: If you saved the return value of each of these lines (which is a netCDF4.Variable object), you could use them in the code below and it would be a bit more legible. e.g.
```python
output_var_min = output_netcdf_obj.createVariable('%s_burst_min' % var, "f4", ("TIME",), fill_value=fillvalue)
...
setattr(output_var_min, 'units', var_units)
...
output_var_min[:] = np.ma.masked_invalid(burst_vars[var]['var_min'])
```

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -0,0 +1,397 @@
...
+            setattr(output_netcdf_obj[var], var_att, getattr(input_netcdf_obj[var], var_att))
+            if var_att != 'comment':
+                setattr(output_netcdf_obj['%s_burst_min' % var], var_att, getattr(input_netcdf_obj[var], var_att))
+                setattr(output_netcdf_obj['%s_burst_max' % var], var_att, getattr(input_netcdf_obj[var], var_att))
+                setattr(output_netcdf_obj['%s_burst_sd' % var], var_att, getattr(input_netcdf_obj[var], var_att))
+
+        var_units = get_imos_parameter_info(var, 'units')
+        if var_units:
+            setattr(output_netcdf_obj[var], 'units', var_units)
+            setattr(output_netcdf_obj['%s_burst_min' % var], 'units', var_units)
```
mhidas: Instead of `setattr` you can assign directly to an attribute, e.g. `output_var_min.units = var_units`  (assuming you assigned `output_var_min` as per comment above)

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -0,0 +1,397 @@
...
+        output_netcdf_obj.createVariable('%s_burst_max' % var, "f4", ("TIME",), fill_value=fillvalue)
+        output_netcdf_obj.createVariable('%s_burst_sd' % var, "f4", ("TIME",), fill_value=fillvalue)
+        output_netcdf_obj.createVariable('%s_num_obs' % var, "i4", ("TIME",), fill_value=fillvalue)
+
+        # set up 'bonus' var att from original FV01 file to FV02
+        input_var_object   = input_netcdf_obj[var]
+        input_var_list_att = input_var_object.__dict__.keys()
+        var_att_disposable = ['name', 'standard_name', 'long_name', 'valid_min', \
+                              'valid_max', '_FillValue', 'ancillary_variables', \
+                              'axis', 'units', 'ChunkSize']
```
mhidas: Why get rid of 'standard_name', 'valid_min', 'valid_max' and units?? Surely these should be the same in the output file as the input file, no?

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -0,0 +1,397 @@
...
+        output_netcdf_obj['%s_burst_max' % var][:] = np.ma.masked_invalid(burst_vars[var]['var_max'])
+        output_netcdf_obj['%s_burst_sd' % var][:]  = np.ma.masked_invalid(burst_vars[var]['var_sd'])
+        output_netcdf_obj['%s_num_obs' % var][:]   = np.ma.masked_invalid(burst_vars[var]['var_num_obs'])
+
+    # add gatts and variable attributes as stored in config files
+    conf_file_generic = os.path.join(os.path.dirname(__file__), 'generate_nc_file_att')
+    generate_netcdf_att(output_netcdf_obj, conf_file_generic, conf_file_point_of_truth = True)
+
+    if instrument_type == 'WQM':
+        conf_file = os.path.join(os.path.dirname(__file__), 'generate_wqm_nc_file_gatt')
```
mhidas: You shouldn't need these anymore.

ANMN/burst_averaged_product/burst_average.py
lbesnard: because the point of truth is taken from this file ```lib/python/imosParameters.txt``` afterwards. I can explain more orally if you'd like

ANMN/burst_averaged_product/burst_average.py
ggalibert: @lbesnard when adding featureType you need to declare a TIMESERIES scalar variable with a specific role attribute. See WQM files processed by a recent toolbox > 2.4. Ex.: http://thredds.aodn.org.au/thredds/dodsC/IMOS/ANMN/NRS/NRSMAI/Biogeochem_timeseries/IMOS_ANMN-NRS_KOSTUZ_20151206T185854Z_NRSMAI-SubSurface_FV01_NRSMAI-SubSurface-24-2015-11-WQM-85_END-20160512T035944Z_C-20160523T032107Z.nc.html

ANMN/burst_averaged_product/burst_average.py
ggalibert: As a side-comment the toolbox outputs LAT and LON in double. In Single precision, the minimum precision available for numbers ranging from -180 to 180 is almost 5 decimal digits (2^-16 = 1.5259e-05 for 2^8 = 256. See http://stackoverflow.com/questions/872544/precision-of-floating-point) which corresponds to ~1m precision at the equator.

Modern GPS usually have a sub-metric precision hence my choice to always output LAT and LON in Double. It would be easier and consistent to keep this variable in Double.

ANMN/burst_averaged_product/burst_average.py
ggalibert: @lbesnard , not a bad idea providing we are all using this version of imosParameters.txt!

I have a version (originally) here : https://github.com/aodn/imos-toolbox/blob/master/IMOS/imosParameters.txt

ANMN/burst_averaged_product/burst_average.py
mhidas: Ah, sorry, I thought we had decided single precision was enough... Keep them double.

ANMN/burst_averaged_product/burstproduct/burst_average_product.py
ggalibert: Sounds a bit dangerous to me to assume a default burst duration. This info, if missing, should not be too difficult to infer from the data.

lbesnard: for this bit i did as Monique used to:
https://github.com/aodn/data-services/blob/master/ANMN/burst_averaged_product/burstproduct/singleCTDburstavproduct.m#L80

ANMN/burst_averaged_product/burst_average.py
mhidas: @ggalibert My point was that we're producing a new file based on the input file, and we're not applying any unit conversion, so at least the output standard_name and units should be the same as in the input file. If that doesn't match what's in imosParameters.txt, that's too bad. Still better than having input temperature in Kelvin and blindly re-labelling it as Celsius!

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -0,0 +1,414 @@
...
+    tmp_netcdf_dir   = tempfile.mkdtemp()
+
+    output_netcdf_file_path = os.path.join(tmp_netcdf_dir, generate_netcdf_burst_filename(input_netcdf_file_path, burst_vars))
+    output_netcdf_obj       = Dataset(output_netcdf_file_path, "w", format="NETCDF4")
+
+    # read gatts from input, add them to output. Some gatts will be
+    # overwritten
+    input_gatts     = input_netcdf_obj.__dict__.keys()
+    gatt_to_dispose = ['author', 'toolbox_input_file', 'file_version_quality_control', 'quality_control_set',
+                       'history', 'CoordSysBuilder_', 'netcdf_filename', 'metadata', 'featureType',
```
mhidas: Please don't delete the 'history' attribute. Just append to it.

ANMN/burst_averaged_product/burst_average.py
ggalibert: If you want to make sure standard_name and units are consistent within input and output files then why don't you leave them as they are without looking at imosParameters.txt??!!

> Still better than having input temperature in Kelvin and blindly re-labelling it as Celsius!

This is exactly what could happen if you look at a version of imosParameters.txt that is not as old as the input file...

mhidas: Unfortunately the version of imosParameters.txt used here is actually a fair bit behind what's in the Toolbox. You've added a few extra columns since I last copied it, so the parsing code needs to be updated. I'll do that.

ggalibert: The more I think about this the more I feel it is a bad idea trying to override the info in the attributes of the input file by what can be found in whichever version of imosParameters.txt. Hope it makes sense.

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -0,0 +1,414 @@
...
+                              '_FillValue', 'ancillary_variables', \
+                              'axis', 'ChunkSize', 'coordinates']
+        for var_att in [att for att in input_var_list_att if att not in var_att_disposable]:
+            setattr(output_netcdf_obj[var], var_att, getattr(input_netcdf_obj[var], var_att))
+            if var_att != 'comment':
+                setattr(output_var_min, var_att, getattr(input_netcdf_obj[var], var_att))
+                setattr(output_var_max, var_att, getattr(input_netcdf_obj[var], var_att))
+                setattr(output_var_sd, var_att, getattr(input_netcdf_obj[var], var_att))
+
+        var_units = get_imos_parameter_info(var, 'units')
```
mhidas: So in light of the conversation with @ggalibert , this is the bit you should definitely not do. The previous loop will copy the units from the input file.

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -0,0 +1,414 @@
...
+        if var_units:
+            output_var_mean.units = var_units
+            output_var_min.units  = var_units
+            output_var_max.units  = var_units
+            output_var_sd.units   = var_units
+
+        output_var_num_obs.units = "1"
+
+        if get_imos_parameter_info(var, 'standard_name'):
```
mhidas: Same here. Use standard_name from input file, not imosParameters.txt

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -0,0 +1,414 @@
...
+            output_var_num_obs.standard_name = "%s number_of_observations" % get_imos_parameter_info(var, 'standard_name')
+
+        if var != 'DEPTH':
+            output_var_mean.coordinates = "TIME LATITUDE LONGITUDE DEPTH"
+        else:
+            attname = 'positive'
```
mhidas: This has already been done in the loop above! No need to do it again.

ANMN/burst_averaged_product/burst_average.py
lbesnard: up to you. I'm doing it but some files won't pass the checker anymore. You get some weird units such as 'degrees_Celsius'

ggalibert: That's fine, you shouldn't be anticipating for problems (and possibly introduce new ones). The input file should be passing the checker. As soon as it does then the product will pass it too.

lbesnard:
@mhidas @ggalibert followed your comments

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -25,15 +25,13 @@ def get_input_file_rel_path(input_netcdf_file_path):
...
-    return file_path
+    return MooringFileClassifier.dest_path(input_netcdf_file_path)
 
 def get_burst_duration(netcdf_file_obj):
     """
     returns the burst duration as float defined in the NetCDF file if exists
     """
-    burst_duration  = getattr(netcdf_file_obj, 'instrument_burst_duration', 60.0)
-    return burst_duration
+    return getattr(netcdf_file_obj, 'instrument_burst_duration', 60.0)
```
ggalibert: This default value is dodgy to me. Especially since this is being used later to identify bursts.

There is/will be cases when it is not 60sec and clearly this can be inferred from the input data itself.

lbesnard: Next commit gets rid of this. It changes slightly the product from Moniques as now there could be more bursts found. I'll do more testing/comparaison

lbesnard:
@ggalibert @mhidas feel free to merge this. it's rebased, and we can re-iterate over it if things have to be modified

PR #465, opened 2016-05-24
A few fixes in IMOSnetCDF.py

mhidas:
* Allow coordinate variables (TIME, LATITUDE, LONGITUDE) to be scalars.
* Improve handling  of variable data types and fill values.
* Create correct filenames for FV02 and FV03 files

These changes are needed for generating the Maria Island timeseries product (PR coming soon). Hopefully they won't affect anything else.

### DIFF
--- a/lib/python/IMOSnetCDF.py
+++ b/lib/python/IMOSnetCDF.py
@@ -120,16 +120,14 @@ class IMOSnetCDFFile(object):
         Create a new variable in the file.
         Returns an IMOSNetCDFVariable object.
         """
-        # if fill_value not given, use default
-        if not fill_value:
-            try:
-                fill_value = self.attributes[name].pop('_FillValue', None)
-            except: pass
 
-        # override data type with default for variable
-        try:
-            dtype = self.attributes[name].pop('__data_type', None)
-        except: pass
+        if name in self.attributes:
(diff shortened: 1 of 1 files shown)


lib/python/IMOSnetCDF.py
```diff
@@ -175,20 +173,20 @@ def updateAttributes(self):
...
-            self.geospatial_lat_max = max(lat[:])
+            self.geospatial_lat_min = lat[:].min()
+            self.geospatial_lat_max = lat[:].max()
 
         # LONGITUDE
         if self.variables.has_key('LONGITUDE'):
             lon = self.variables['LONGITUDE']
-            self.geospatial_lon_min = min(lon[:])
-            self.geospatial_lon_max = max(lon[:])
+            self.geospatial_lon_min = lon[:].min()
```
lbesnard: I had an issue lately with this. Some netcdf files have LATITUDE and LONGITUDE set as dimensionless var. And because of this, they have a shape of 0. and min and max can't be performed on 0-d array. So maybe worth handling this type of scenario as well

mhidas: Yes, that's exactly what this change is doing. This will work in those cases.

mhidas:
@lbesnard Care to merge this?

PR #467, opened 2016-05-26
SOOP_BA: fix bug and add processing of new type of files

bpasquer:
- delete original netcdf file after pushing  processed tmp_nc version to S3
 - add processing of transit csv files

### DIFF
--- a/SOOP/SOOP_BA/incoming_handler.sh
+++ b/SOOP/SOOP_BA/incoming_handler.sh
@@ -42,14 +42,12 @@ delete_previous_versions() {
 
     if has_extension $file "nc"; then
         del_function='s3_del'
-    elif has_extension $file "png"; then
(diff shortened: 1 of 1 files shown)


SOOP/SOOP_BA/incoming_handler.sh
```diff
@@ -42,14 +42,13 @@ delete_previous_versions() {
 
     if has_extension $file "nc"; then
         del_function='s3_del'
-    elif has_extension $file "png"; then
+    elif has_extension $file "png" ||  has_extension $file "csv"; then
         local file_basename=`basename $file`
```
mhidas: `file_basename` is the same as `basename_file` above (line 35). Probably don't need both?

bpasquer:
You're right @mhidas i removed file_basename

PR #471, opened 2016-05-31
DELETE depreciated code

lbesnard:
* AATAMS DM
 * SOOP aggregation
 * SRSALT
 * SRS sst aws migration


mhidas:
@lbesnard I think the word you're after is [deprecated](https://en.wikipedia.org/wiki/Deprecation), or better still *obsolete*.

mhidas:
It's not easy to review this when everything is being deleted. I see that it's mostly Matlab code, so that makes sense. The unit tests pass, so I assume it's all good.

While we're at it, I've added some more deletions of my own.

lbesnard:
yeap, fair enough. It's mostly very old code/code with no cron that I wrote. Thanks for adding stuff :+1:  
@bpasquer  @ggalibert feel free to add more obsolete code to delete

ggalibert:
I've removed some old ACORN code too. Do we want to remove old burst average matlab code?

lbesnard:
> I've removed some old ACORN code too. Do we want to remove old burst average matlab code?

it's already part of the burst average PR

lbesnard:
Thanks all, I squashed all the commits into one

PR #475, opened 2016-06-03
Declare I and J variables for CODAR hourly average files in order to …

ggalibert:
…be CF compliant.


ggalibert:
Please do not merge yet

PR #479, opened 2016-06-08
remove unused JAVA classes

lbesnard:


lbesnard:
@mhidas with this PR, the repo is only a few MB (apart of history)

@lwgordonimos I guess that generally speaking, repos can be clone with a depth equal to 1 on the different nodes. Prob faster to do a provision on a node from scratch. what do you think ?

mhidas:
:+1: nice one @lbesnard 
Is it possible to remove these from the history as well?

lwgordonimos:
@lbesnard I would think it would be faster yes. I can't think of a reason why you would ever need full revision history in the clone when it is only ever going to use the HEAD in the master branch for provisioning.

lbesnard:
> Is it possible to remove these from the history as well?

hum i don't really want to take this responsibility, and do a push-force on master. I don't think it's worth it

lwgordonimos:
No definitely not, not over a matter of 20MB.

PR #483, opened 2016-06-14
 SOOP_CO2: add basic report to uploader

bpasquer:


SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -26,11 +26,15 @@ is_future_reef_map_file() {
...
     log_info "Handling SOOP CO2 zip file '$file'"
+    local recipient=`get_uploader_email $file`
 
     local tmp_dir=`mktemp -d`
     chmod a+rx $tmp_dir
     local tmp_zip_manifest=`mktemp`
 
+    echo "" | notify_by_email $BACKUP_RECIPIENT "Processing new underway CO2 zip file '$file'"
+    echo "" | notify_by_email $recipient "Processing new underway CO2 zip file '$file'"
+
```
mhidas: It would make sense to group these lines together with the `log_info` line above. The lines currently between them could come after these.

SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -79,6 +81,9 @@ main() {
         s3_put_no_index $extracted_file $path/$file_basename
     done
 
+    echo "" | notify_by_email $recipient "Successfully published SOOP_CO2 file '$path' "
+    echo "" | notify_by_email $BACKUP_RECIPIENT "Successfully published SOOP_CO2 file '$path'"
+
```
mhidas: (General comment) Sending two emails to the uploader for each file (+2 to yourself) is ok if only a few files are uploaded at a time. In the long run it would be much better to have a way to generate a single report summarising all relevant events during the incoming handling, and just send it once at the end, whether the process terminated happily or with an error.

bpasquer:
I gathered log lines at the beginning of the script as suggested .

SOOP/SOOP_CO2/incoming_handler.sh
bpasquer: I agree, it is ok cause i know i only get a few files. And again yes, having a single report at the end of the process is the way to go.

PR #486, opened 2016-06-15
WERA hourly files have their GDOP test updated so that suspicious cas…

ggalibert:
…e is also flagged 4.


ggalibert:
Please do not merge yet

Issue #493, opened 2016-06-17
ANMN burst average product might be out of sync with original FV01 in terms of metadata


ggalibert:
As I was telling Loz, If we rebuild a NetCDF from scratch looking at a lookup table for variable attributes and for a specific NetCDF structure (dimensions, extra scalar variables, global attributes, variable attributes), the day we update the content, the structure of the FV01 file or the nature of the CF/IMOS checks, the resulted FV02 will not be consistent with the FV01 files or won't pass the new checks. The lookup table will need updating and the burst average code will need to be modified to adapt to a new file structure or new check.

We don't need to worry about historical files that don't pass checks. They should be re-uploaded or fixed once and for all. This burst average process should be as agnostic as possible from what makes a file pass or not the checker. All it should do is process files that have just passed the checker in order to produce files that will automatically pass it again.

A simpler and more generic approach consists in taking the FV01 file as it is for starting building a new FV02.

For example, if we have:
```dimensions:
        TIME = 250840 ;
variables:
        double TIME(TIME) ;
                TIME:standard_name = "time" ;
                TIME:long_name = "time" ;
                TIME:units = "days since 1950-01-01 00:00:00 UTC" ;
                TIME:calendar = "gregorian" ;
                TIME:axis = "T" ;
                TIME:valid_min = 0. ;
                TIME:valid_max = 90000. ;
        double LATITUDE ;
                LATITUDE:standard_name = "latitude" ;
                LATITUDE:long_name = "latitude" ;
                LATITUDE:units = "degrees_north" ;
                LATITUDE:axis = "Y" ;
                LATITUDE:reference_datum = "geographical coordinates, WGS84 projection" ;
                LATITUDE:valid_min = -90. ;
                LATITUDE:valid_max = 90. ;
        double LONGITUDE ;
                LONGITUDE:standard_name = "longitude" ;
                LONGITUDE:long_name = "longitude" ;
                LONGITUDE:units = "degrees_east" ;
                LONGITUDE:axis = "X" ;
                LONGITUDE:reference_datum = "geographical coordinates, WGS84 projection" ;
                LONGITUDE:valid_min = -180. ;
                LONGITUDE:valid_max = 180. ;
        float NOMINAL_DEPTH ;
                NOMINAL_DEPTH:standard_name = "depth" ;
                NOMINAL_DEPTH:long_name = "nominal depth" ;
                NOMINAL_DEPTH:units = "metres" ;
                NOMINAL_DEPTH:axis = "Z" ;
[...]

PR #495, opened 2016-06-17
Burst average improvements

lbesnard:
* clean code with the assumption that the FV01 input file is compliant


lbesnard:
@ggalibert in response to https://github.com/aodn/data-services/issues/493

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -148,12 +148,24 @@ def list_var_to_average(netcdf_file_obj):
...
 
-    for var_to_remove in ['TIME', 'LATITUDE', 'LONGITUDE', 'NOMINAL_DEPTH', 'VOLT', 'TIMESERIES', 'SSPD', 'CNDC']:
-        if var_to_remove in var_list:
-            var_list.remove(var_to_remove)
+    var_to_remove = []
+    for varname in var_list:
+        if 'TIME' not in netcdf_file_obj.variables[varname].dimensions:
+            var_to_remove.append(varname)
+
+    var_to_remove.extend(('TIME', 'VOLT', 'SSPD', 'CNDC'))
```
lbesnard: apart of 'TIME', Monique used to get rid of these variables as well

lbesnard:
[IMOS_ANMN-NRS_KOSTUZ_20151207T024417Z_NRSMAI_FV02_NRSMAI-SubSurface-24-2015-11-WQM-85-burst-averaged_END-20160507T235914Z_C-20160617T134729Z.nc.zip](https://github.com/aodn/data-services/files/319758/IMOS_ANMN-NRS_KOSTUZ_20151207T024417Z_NRSMAI_FV02_NRSMAI-SubSurface-24-2015-11-WQM-85-burst-averaged_END-20160507T235914Z_C-20160617T134729Z.nc.zip)

ANMN/burst_averaged_product/burst_average.py
ggalibert: Ok to remove these variables, except TIME. TIME should stay as is. Only data and comment attribute should be updated.

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -148,12 +148,24 @@ def list_var_to_average(netcdf_file_obj):
...
     var_list = [x for x in var_list if not x.endswith('_quality_control')]
 
-    for var_to_remove in ['TIME', 'LATITUDE', 'LONGITUDE', 'NOMINAL_DEPTH', 'VOLT', 'TIMESERIES', 'SSPD', 'CNDC']:
-        if var_to_remove in var_list:
-            var_list.remove(var_to_remove)
+    var_to_remove = []
+    for varname in var_list:
+        if 'TIME' not in netcdf_file_obj.variables[varname].dimensions:
+            var_to_remove.append(varname)
+
```
ggalibert: This is not what I suggested to do:
> We want to keep everything as it is, except:
-we get rid of any '_quality_control' variable.
-we duplicate any variable PARAM that is a function of TIME so that we have: PARAM, PARAM_min, PARAM_max, PARAM_sdv and PARAM_num_obs. These variables will automatically inherit the data and attributes from PARAM. Apply mean or median, min, max, standard_dev and count to the data of each of them respectively (flags are taken into account to only do this on "good" data) and update the relevant attributes accordingly, add cell_methods, etc...
-TIME is treated slightly differently since it doesn't have a TIME_min, TIME_max, etc...
-for global attributes, most of them will still be relevant, some will need to be deleted, some updated.

So any variable that is not a function of TIME is kept.

lbesnard: thats the case. There is a new function listing them

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -148,12 +148,24 @@ def list_var_to_average(netcdf_file_obj):
...
+    var_to_remove.extend(('TIME', 'VOLT', 'SSPD', 'CNDC'))
+    var_list = [x for x in var_list if x not in var_to_remove]
 
     return var_list
 
+def list_dimensionless_var(netcdf_file_obj):
+    var_list    = netcdf_file_obj.variables.keys()
```
lbesnard: @ggalibert see this function. Then see ligne 288 where they get created

lbesnard:
@ggalibert, I will probably look at the xarray way in another PR as a lot would have to be modified. But If you 're happy, this can be merged I reckon

PR #496, opened 2016-06-20
AATAMS SATTAG DM

lbesnard:
* dont run lftp if the $ERROR_DIR is not empty

### DIFF
--- a/AATAMS/AATAMS_sattag_dm/lftp_sync.sh
+++ b/AATAMS/AATAMS_sattag_dm/lftp_sync.sh
@@ -23,10 +23,15 @@ sync_files() {
 # main
 main() {
     mkdir -p $ZIPPED_DIR
-    local tmp_lftp_output_file=`mktemp`
-    sync_files $tmp_lftp_output_file
 
-    mv $tmp_lftp_output_file $INCOMING_DIR/AATAMS/AATAMS_SATTAG_DM/aatams_sattag_dm_lftp.`date +%Y%m%d-%H%M%S`.log
+    # look for empty dir
+    if [ "$(ls -A $ERROR_DIR/AATAMS_SATTAG_DM)" ]; then
+         echo "Unable to run lftp as $ERROR_DIR/AATAMS_SATTAG_DM is not empty"
+    else
+        local tmp_lftp_output_file=`mktemp`
+        sync_files $tmp_lftp_output_file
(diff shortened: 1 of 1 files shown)


lbesnard:
This should avoid to have too many lftp log files to process in incoming in case something is wrong with the harvester.

AATAMS/AATAMS_sattag_dm/lftp_sync.sh
```diff
@@ -23,10 +23,15 @@ sync_files() {
 # main
 main() {
     mkdir -p $ZIPPED_DIR
-    local tmp_lftp_output_file=`mktemp`
-    sync_files $tmp_lftp_output_file
 
-    mv $tmp_lftp_output_file $INCOMING_DIR/AATAMS/AATAMS_SATTAG_DM/aatams_sattag_dm_lftp.`date +%Y%m%d-%H%M%S`.log
+    # look for empty dir
+    if [ "$(ls -A $ERROR_DIR/AATAMS_SATTAG_DM)" ]; then
+         echo "Unable to run lftp as $ERROR_DIR/AATAMS_SATTAG_DM is not empty"
```
lwgordonimos: Have no specific knowledge of this specific script, but if the script generates any output, Cron should email it to the MAILTO address. @lbesnard is this your expectation?

https://github.com/aodn/data-services/blob/master/cron.d/AATAMS_SATTAG_DM

lbesnard: Currently, the lftp/rsync runs as set up in the cron (similar thing for both AATAMS SATTAG DM and ARGO). It adds/removes files to/from $WIP_DIR. Every 'additions' and 'deletion' is written in a manifest file which gets pushed to the incoming directory.

In the case the manifest file doesn't get processed (this was the case with ARGO), more manifest files are still generated. Each manifest file will be unique, containing different additions and deletions. In the case these manifest files landed in the ERROR_DIR, it is impossible to run them again, because some files have disappeared physically from the $WIP_DIR ($WIP_DIR is a FS and the increment is done via the manifest).

It's exactly what happens with ARGO. I have to look now at the manifest files kind of manually to generate semi-manually a new manifest file combining the last 2....

So this PR prevents the cron to run in case a file has landed to ERROR_DIR.

makes sense ?

lbesnard: > Where will this message go? Will you know about it?

I'll know about it because NAGIOS will always show 1 file in ERROR_DIR

mhidas: Ok, fair enough.

PR #497, opened 2016-06-20
IMOS checker plugin updates

mhidas:
* Update README
* Add metadata attributes to IMOSCheck object
* Include test data as CDL instead of NetCDF for easier versioning.


ggalibert:
> * Include test data as CDL instead of NetCDF for easier versioning.

Great idea!

Issue #502, opened 2016-06-21
ANMN burst average new global attribute product_processing_description


ggalibert:
Is [this](https://github.com/aodn/data-services/blob/7e21d8791d5494ff3473f605c717ced749197ef9/ANMN/burst_averaged_product/burst_average.py#L388) new global attribute `product_processing_description` really useful?

I can understand it could be good to trace where this FV02 file comes from but this reference is half addressing the question since this particular file in master can have a myriad of different versions. Either you keep track of a very identified version of the code or you don't.

Issue #501, opened 2016-06-21
ANMN burst average list_var_to_average improvement


ggalibert:
For the function [list_var_to_average](https://github.com/aodn/data-services/blob/7e21d8791d5494ff3473f605c717ced749197ef9/ANMN/burst_averaged_product/burst_average.py#L144), In order to list variables that needs to be averaged (function is relevantly named `list_var_to_average`), I would have listed the specific parameters to averaged rather than the one we don't want to average. Indeed, in case there is a new parameter that is introduced, we don't risk to introduce it in FV02 without knowing. The day we want to add a parameter, at least it is a conscious choice.

Issue #500, opened 2016-06-21
ANMN burst average clean_burst_vars question


ggalibert:
What is the purpose of this [function](https://github.com/aodn/data-services/blob/7e21d8791d5494ff3473f605c717ced749197ef9/ANMN/burst_averaged_product/burst_average.py#L71)? Would be good to comment better on this.

lbesnard:
basically to get rid of timestamps at the start and end of a FV02 file when all FV02 variables have a NaN value

ggalibert:
Can we rename this function with something like `trim_timestamps_burst_vars` and add your comment above to the code? I didn't find it clear as it is commented now.

Issue #499, opened 2016-06-21
ANMN burst average input flag question


ggalibert:
Does [this](https://github.com/aodn/data-services/blob/7e21d8791d5494ff3473f605c717ced749197ef9/ANMN/burst_averaged_product/burst_average.py#L31) come from Monique's original code? Can you explain the rational behind? I'm not sure to understand... I would have removed anything that is not flagged 0, 1 or 2.

lbesnard:
see the 4 lines above Monique's code https://github.com/aodn/data-services/blob/f90a2b417fbdc0218de9b07adcf7bb8b1a0d9581/ANMN/burst_averaged_product/burstproduct/singleWQMburstavproduct.m#L117

ggalibert:
Ok, I can see she removes only flags 4 for 'TURB','FLU2','CPHL','CHLR','CHLU','CHLF' and removes flags 3 and 4 for the rest. I reckon this was based on the fact that the rate of change test which would provide flag 3 is not relevant for bio-optical sensors. Anyway, once again we shouldn't try to interfer with the original content of the FV01 file and assume the QC has been properly performed. Conclusion:  we should remove anything that has a flag >= 3 before averaging.

PR #503, opened 2016-06-21
Burst Average Improvements

lbesnard:
* remove now all flags >= 3 for ALL variables
 * add SPEC_CNDC in list of variables not to process
 * improve product_processing_description gatt and add github url
   generation to util.py
 * rename function to trim_timestamps_burst_vars

### DIFF
--- a/ANMN/burst_averaged_product/burst_average.py
+++ b/ANMN/burst_averaged_product/burst_average.py
@@ -9,8 +9,8 @@ from file_classifier import MooringFileClassifier
 from generate_netcdf_att import generate_netcdf_att, get_imos_parameter_info
 from math import isnan
 from netCDF4 import Dataset, num2date, date2num
+from util import get_git_revision_script_url
 import argparse
-import inspect
 import numpy as np
 import os
 import pandas as pd
@@ -28,29 +28,6 @@ def get_input_file_rel_path(input_netcdf_file_path):
     """
     return MooringFileClassifier.dest_path(input_netcdf_file_path)
 
-def min_bad_qc_flag(varname):
-    """returns the min IMOS IODE qc flag value to get rid of per variable"""
-    default_min_qc_flag_value = 3
-    dict_var = {
-        'CHLF':     4,
-        'CHLR':     4,
-        'CHLU':     4,
-        'CPHL':     4,
-        'DOX1_2':   3,
-        'DOX2':     3,
-        'FLU2':     4,
-        'FLNTU':    4,
-        'PRES_REL': 4,
-        'PSAL':     4,
-        'TEMP':     4,
-        'TURB':     4,
-    }
-
-    if varname in dict_var.keys():
-        return dict_var[varname]
-    else:
-        return default_min_qc_flag_value
-
 def create_burst_average_var(netcdf_file_obj):
     """
     create burst data from all vars available in netcdf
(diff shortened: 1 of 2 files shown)


lbesnard:
@ggalibert 
amend 
https://github.com/aodn/data-services/issues/499
https://github.com/aodn/data-services/issues/500
https://github.com/aodn/data-services/issues/501
https://github.com/aodn/data-services/issues/502

ggalibert:
:+1:

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -381,11 +360,7 @@ def create_burst_average_netcdf(input_netcdf_file_path, output_dir):
 
     # add values to variables
     output_netcdf_obj['TIME'][:]          = np.ma.masked_invalid(time_burst_vals)
-
-    product_processing_description = inspect.getfile(inspect.currentframe())
-    script_dir_path = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
-    product_processing_description = os.path.join(script_dir_path[script_dir_path.index('data-services') + len('data-services') + 1:], os.path.basename(product_processing_description))
-    output_netcdf_obj.product_processing_description = 'Product created by www.github.com/aodn/data-services/tree/master/%s' % product_processing_description
+    output_netcdf_obj.product_processing_description = 'Product created with %s' % get_git_revision_script_url(os.path.realpath(__file__))
```
mhidas: This should go into one of the standard attributes `lineage` or `history`.

lib/python/util.py
```diff
@@ -34,3 +34,21 @@ def list_files_recursively(dir, pattern):
             matches.append(os.path.join(root, filename))
 
     return matches
+
+def get_git_revision_script_url(file_path):
+    import os
+    import subprocess
```
mhidas: According to [PEP 8](https://www.python.org/dev/peps/pep-0008/#id23) all imports should be at the top of a file.  I guess there may be cases when you only want to import into a function's local namespace, but I don't think that is the case here.

lbesnard: @mhidas agree with you, however the reason I did this is written in the above lines https://github.com/aodn/data-services/blob/00c65c91cac4b748cb67b9b0ac1264b27c56be3d/lib/python/util.py#L5

Since it will hold various functions, In case some rewriting would be required, I decided to make the functions a bit more independent. what do you think ?

lib/python/util.py
```diff
@@ -34,3 +34,21 @@ def list_files_recursively(dir, pattern):
...
+    """
+    file_path is the local file path from a git repo
+    returns the github url with the hash value of the current HEAD
+    """
+    curr_dir = os.getcwd()
+    os.chdir(os.path.dirname(file_path)) # need to chg dir to run git commands
+
+    repo_username_name = (subprocess.check_output(['git', 'config', '--get', 'remote.origin.url']).strip()).split(':')[1]
```
mhidas: Don't need the extra brackets around `subprocess.check_output().strip()`

lib/python/util.py
```diff
@@ -34,3 +34,21 @@ def list_files_recursively(dir, pattern):
...
+    repo_name          = repo_username_name.split('/')[1]
+    hash_val           = subprocess.check_output(['git', 'rev-parse', 'HEAD']).strip()
+    script_rel_path    = file_path[file_path.index(repo_name) + len(repo_name) + 1:]
```
mhidas: @lbesnard  Have you tested this? As far as I can tell, `repo_name` will end up something like 'data-services.git', but file_path will only contain 'data-services', so this should raise an exception.

lbesnard: @mhidas yeap tested.
```bash
git config --get remote.origin.url
git@github.com:aodn/data-services
```
so repo_name won't have a .git in its  name

lbesnard: actually the same command does't return the same thing for all machines/repos. looking at it now

lwgordonimos:
Have you considered using https://github.com/gitpython-developers/GitPython instead of subprocess? Subprocess is the devil and this module is in PyPI...

lbesnard:
@lwgordonimos yes, and the issue I have found is that you have to give the root of the git folder to initialise anything, which defeats the purpose. But if you find out something, i'd be more than happy

mhidas:
But we do know the root git folder. It's in the environment variable `$DATA_SERVICES_DIR`, or you can start with `file_path` and go up the dir tree until you find a .git directory.

lbesnard:
> and go up the dir tree until you find a .git directory.
that's what I was about to do actually

lib/python/util.py
mhidas: Ok.

PR #505, opened 2016-06-22
Burst average Improvements

lbesnard:
* improve github url creation in util.py (no more subprocess)
 * add gitpython in requirements
 * use fillValue and dtype of input file for var creation of output
 * mv github url att to lineage att
 * clean lineage att talking about different QC flag values
 * FillValue for i4 num_obs var takes default i4 fillvalue


lbesnard:
@lwgordonimos no more subprocess for the github link creation
@mhidas followed your comments
FYI @ggalibert

ANMN/burst_averaged_product/generate_nc_file_att
```diff
@@ -2,4 +2,4 @@
...
+lineage                 = The data array in this file has been created by binning (averaging) raw burst data. Each array value is the arithmetic mean of a burst. Length and frequency of the underlying bursts are declared in the global attributes instrument_burst_duration and instrument_burst_interval respectively. Out-of-water data is excised before binning. Points flagged by IMOS QC as 3 and above are excluded from the burst average: this includes data values outside of a valid range, data depth outside of a valid range, and data values in a series of consecutive identical values whose length exceeds a maximum. These flags are produced by the IMOS rate of change test: this is a test to detect spikes. This sort of test is not used to exclude any optical sensor data, because understanding of behaviour of these parameters at this time scale is still in its early stages. Burst means are calculated from remaining points. Any bursts with no remaining good points after exclusions are assigned FillValue (see variable attributes for value of FillValue).
```
ggalibert: I didn't have a look at this so far. Sorry but this will need to be removed:

>: this includes data values outside of a valid range, data depth outside of a valid range, and data values in a series of consecutive identical values whose length exceeds a maximum. These flags are produced by the IMOS rate of change test: this is a test to detect spikes. This sort of test is not used to exclude any optical sensor data, because understanding of behaviour of these parameters at this time scale is still in its early stages.

Also, it wouldn't harm to say again here that the time stamp lays in the middle of the burst. Can you please check this is the case? Could also be `(tn - t1)/2` with `t1` the time of the first good data in burst and `tn` the time of the last good data in burst.

ggalibert:
It hink it's a comment @mhidas already made but I find it easier/neater to review a PR when there is one commit per distinct task covered.

ANMN/burst_averaged_product/generate_nc_file_att
lbesnard: > Can you please check this is the case?
it is. https://github.com/aodn/data-services/blob/master/ANMN/burst_averaged_product/burst_average.py#L174

lbesnard:
@ggalibert fixed lineage. can rebase if you re happy

lib/python/util.py
```diff
@@ -36,19 +36,35 @@ def list_files_recursively(dir, pattern):
...
+        curr_dir = os.path.abspath(os.path.join(curr_dir, os.pardir))
+        if curr_dir == os.path.sep:
+            raise Exception('Cannot find recursively a .git in parent directories')
+
+    repo_name      = os.path.basename(curr_dir)
+    repo           = Repo(curr_dir)
+    remote_org_url = repo.remotes.origin.url
+
+    pattern = 'git@github.com:(.*)/%s' % repo_name
+    regroup = re.search(pattern, remote_org_url)
```
mhidas: Not sure why you bother extracting `organisation` out of `remote_org_url`, when you just stick it back in front of the `repo_name` for the output, and don't use it anywhere else. All you need to do is replace "git@github.com:" with "www.github.com/" in `remote_org_url` (and drop a possible ".git" at the end), and you have the start of your return value.

lbesnard: > for the output, and don't use it anywhere else.

I don't understand what you mean, this is the point, to only use it to put it in front of the repo name.

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -259,23 +259,23 @@ def create_burst_average_netcdf(input_netcdf_file_path, output_dir):
...
-        output_netcdf_obj.createVariable(var, input_netcdf_obj[var].dtype, ())
+        output_netcdf_obj.createVariable(var, input_netcdf_obj[var].dtype)
         output_netcdf_obj[var][:] = input_netcdf_obj[var][:]
 
     for var in burst_vars.keys():
-        fillvalue = get_imos_parameter_info(var, '_FillValue')
-        if fillvalue == []:
-            fillvalue = 999999.0
+        var_dtype = input_netcdf_obj[var].dtype
+        fillvalue = getattr(input_netcdf_obj[var], '_FillValue', default_fillvals[np.dtype(var_dtype).str[1:]])
```
mhidas: You can just use `None` as the default value. Calling `createVariable` with `fillvalue=None` has the same effect as what you're doing here.

lib/python/util.py
mhidas: I mean it is already in front of the repo name in `remote_org_url`, so no need to separate it.

lbesnard: > I mean it is already in front of the repo name in remote_org_url, so no need to separate it.

well its just a different way. This way you don't have to handle the case of removing .git or not. Both are legit i guess

lbesnard:
@mhidas done

lwgordonimos:
Git function could be condensed further with something like this... less string manipulation/slicing/dicing etc.:

```python
def get_git_revision_script_url(file_path):
    """
    file_path is the local file path in a github repo
    returns the github url with the hash value of the current HEAD
    Only handles `git config --get remote.origin.url` output in the form defined
    in pattern variable
    """
    import os
    import re
    from git import Repo

    # this will throw an InvalidGitRepositoryError for a invalid repo, but this is relevant and descriptive, so let it
    # throw (let it throw! Can't hold me back any moooore... wow sorry about that... stupid Frozen...)
    repo = Repo(file_path, search_parent_directories=True)
    hash_val = str(repo.commit('HEAD'))
    script_rel_path = os.path.relpath(file_path, repo.working_tree_dir)

    pattern = '(.*)@(.*):(.*)/(.*)\.git'
    regroup = re.search(pattern, repo.remotes.origin.url)
    try:
        user, host, organisation, repo_name = regroup.group(1, 2, 3, 4)
    except ValueError:
        # the default message is very vague, so rethrow with a more descriptive message
        raise ValueError('Cannot parse remote.origin.url')

    return '{0}/{1}/{2}/blob/{3}/{4}'.format(host, organisation, repo_name, hash_val, script_rel_path)
```

lbesnard:
@lwgordonimos thanks, did what you suggested. However had to get rid of .git in patter as its not always there, and added a repo_name.strip('.git') in case it is

mhidas:
@lbesnard The pattern above will not include ".git" in the `repo_name`. However, it will throw an exception if that bit is missing from the returned url.

You could allow for both cases with the pattern `'(.*)@(.*):(.*)/(.*)$'`, and in that case you would need to remove ".git" from the end of `repo_name`. However, `.strip('.git')` is not what you want. That will remove any of the characters '.', 'g', 'i', or 't' from either end of the string! Instead, you could do something like
```python
re.sub('\.git$', '', repo_name)
```

lbesnard:
@mhidas @ggalibert ready to be merged I guess

ggalibert:
good forme. Please for next PR: do distinct commits for distinct tasks...

Issue #283, opened 2015-12-08
Python  coding style and consistency


mhidas:
Now that more of us are starting to write more Python code, it might be helpful to converge a bit in terms of coding style. In particular, naming conventions for variables, functions, classes and files. It just makes the code a bit more readable, so it's worth putting _a bit_ of effort (not heaps) into it. As [this detailed style guide for Python code](https://www.python.org/dev/peps/pep-0008/) suggests "A style guide is about consistency. Consistency with this style guide is important. Consistency within a project is more important. Consistency within one module or function is most important. "

My own code needs some cleaning up, but first we should agree on what to aim for. I propose we adopt the following conventions as an ideal:
- Use 4 spaces (not tabs) per indentation level, and >=4 spaces for long lines ([examples](https://www.python.org/dev/peps/pep-0008/#indentation))
- Module names (i.e. names of .py files) should be `lower_case`, with underscores if necessary
- Class names should use `CapWords` (capitalise first letter of every word, no underscores). If any word is an abbreviation, capitalise the whole word, e.g. use `ANMNFileClassifier` instead of `AnmnFileClassifier`
- Function names should be `lower_case`, with underscores between words (alternatively, if we decide we have too many in `mixedCase` already, we can stick with that, though IMHO they're less readable. Also we use the lower_case convention in our bash scripts)
- Variable names should be `lower_case`, same as functions
- Use leading underscores for non-public methods and variables
- Constants within a module should be defined at the top and use `UPPER_CASE`
- [Docstrings](https://www.python.org/dev/peps/pep-0008/#documentation-strings) - describe each module, class and function by adding a "triple-double-quoted" string directly after the definition, e.g.

``` python
    def do_something(arg):
        """Do something with arg and return the result.
        If no argument given, do nothing.
        """"
```

These are just suggestions, open for discussion.

PR #506, opened 2016-06-23
Hourly files created with no valid current data are ignored and not p…

ggalibert:
…ushed to S3.

### DIFF
--- a/ACORN/current_generator/acorn_utils.py
+++ b/ACORN/current_generator/acorn_utils.py
@@ -15,7 +15,7 @@ class Enum(set):
             return name
         raise AttributeError
 
-ACORNError = Enum(["NOT_ENOUGH_FILES", "ERROR", "SUCCESS"])
+ACORNError = Enum(["NOT_ENOUGH_FILES", "NO_CURRENT_DATA", "ERROR", "SUCCESS"])
 
 def file_parts(f):
     return f.split(acorn_constants.DELIMITER)
--- a/ACORN/current_generator/acorn_utils.py
+++ b/ACORN/current_generator/acorn_utils.py
@@ -15,7 +15,7 @@ class Enum(set):
             return name
         raise AttributeError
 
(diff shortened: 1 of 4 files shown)


ACORN/current_generator/wera.py
```diff
@@ -66,6 +66,19 @@ def has_enough_radials(radial_file_list):
         return retval
 
     @staticmethod
+    def has_current_data(nc_file):
+        # Return True if netcdf file includes at least 1 current measurement != NaN
+        retval = True
+        F = Dataset(nc_file, mode='r')
+        if F['UCUR'][:].mask.all() & F['VCUR'][:].mask.all():
```
mhidas: It may not make a difference in some cases, but I think you want the boolean operator `and` here.
`&` means *bitwise* AND.

ACORN/current_generator/acorn_utils.py
```diff
@@ -15,7 +15,7 @@ def __getattr__(self, name):
             return name
         raise AttributeError
 
-ACORNError = Enum(["NOT_ENOUGH_FILES", "ERROR", "SUCCESS"])
+ACORNError = Enum(["NOT_ENOUGH_FILES", "NO_CURRENT_DATA", "ERROR", "SUCCESS"])
```
mhidas: According to the [coding conventions](https://github.com/aodn/data-services/issues/283) this module constant should be named `ACORN_ERROR`.
Also, considering "SUCCESS" is one of the options, maybe `ACORN_STATUS` would be a better name.

ggalibert:
Thanks for your feedback @mhidas . If you're happy with newly added commits to address your comments I will squash them into one.

mhidas:
@ggalibert Looks good. Squash away!
Just a general comment though, it looks like there's a bit of code redundancy between wera.py and codar.py that could be refactored at some point.

mhidas:
@ggalibert I let GitHub do the squashing. Hope you don't mind.

PR #517, opened 2016-07-06
fix wrong dotenv python package installed

lbesnard:


lbesnard:
@lwgordonimos I got confused with different python dotenv packages, and put the wrong one in requirements last time. Would be probably cleaner to to a pip uninstall dotenv of the old one

Issue #513, opened 2016-07-04
IOOS checker and plugins new installation makes AIMS scripts fail


lbesnard:
FAIMMS, ANMN NRS DARWIN YONGALA and SOOP TRV scripts runs as CRONTAB everyday. They also use the NETCDF-CHECKER before pushing a new file to the $INCOMING_DIR to prevent bad files to be downloaded and pushed. This used to work with no issues until last tuesday night on both NSP14 and AWS10.

The issue seems to be related to a bad installation of the IOOS CHECKER/plugin system.

On __NSP14__, run ipython, and paste the following lines
```python
import os
import sys
import tempfile
netcdf_file_path='/mnt/imos-test-data/IMOS/FAIMMS/Heron_Island/Relay_Pole_1/sea_water_temperature@1.7m_channel_26/2009/NO_QAQC/IMOS_FAIMMS_T_20090101T000000Z_HIRP1_FV00.nc'
netcdf_checker_path = os.path.dirname(os.path.realpath(os.environ.get('NETCDF_CHECKER')))
sys.path.insert(0, netcdf_checker_path)
import cchecker

tmp_json_checker_output = tempfile.mkstemp()
test='imos'
return_value, errors = cchecker.ComplianceChecker.run_checker(netcdf_file_path, [test] , 'None', 'normal', tmp_json_checker_output[1], 'json')
```

This will fail with the following message :
```python
No valid checkers found for tests 'imos'
---------------------------------------------------------------------------
UnboundLocalError                         Traceback (most recent call last)
/home/lbesnard/data-services/FAIMMS/REALTIME/<ipython-input-17-eeb824a74047> in <module>()
----> 1 return_value, errors = cchecker.ComplianceChecker.run_checker(netcdf_file_path, [test] , 'None', 'normal', tmp_json_checker_output[1], 'json')

/var/lib/netcdf-checker/compliance_checker/runner.py in run_checker(cls, ds_loc, checker_names, verbose, criteria, output_filename, output_format)
     48 
     49         elif output_format == 'json':
---> 50             groups = cls.json_output(cs, score_groups, output_filename, ds_loc, limit)
     51 
     52         else:

/var/lib/netcdf-checker/compliance_checker/runner.py in json_output(cls, cs, score_groups, output_filename, ds_loc, limit)
    121                     cs.json_output(checker, groups, f, ds_loc, limit)
    122 
--> 123         return groups
    124 
    125     @classmethod

[...]

PR #521, opened 2016-07-06
use cchecker Egg

lbesnard:


lbesnard:
@lwgordonimos at least this is cleaner !

lib/aims/realtime_util.py
```diff
@@ -12,6 +12,7 @@
 author Laurent Besnard, laurent.besnard@utas.edu.au
 """
 
+from compliance_checker.runner import import ComplianceChecker
```
lwgordonimos: Noticed duplicated "import"! Also, see https://github.com/aodn/data-services/issues/513#issuecomment-230698585

lbesnard:
@lwgordonimos thanks for your help.

This cannot be tested anymore on PO box though. but should be working fine on rc and prod

PR #526, opened 2016-07-12
Update checker info added to passed files

mhidas:
Now that we're deploying the compliance checker from an egg, the git commit info is no longer easily accessible. We'll just keep the release version of the core code. Also, we need to keep track of the imos plugin version.

### DIFF
--- a/lib/common/netcdf-check.sh
+++ b/lib/common/netcdf-check.sh
@@ -81,25 +81,11 @@ _get_checker_code_dir() {
 }
 export -f _get_checker_code_dir
 
-# return the current commit hash of the checker
-_get_checker_commit_hash() {
-    cd `_get_checker_code_dir` && \
-        git log -1 --format=format:'%H'
-}
-export -f _get_checker_commit_hash
-
-# return the date/time of the current commit of the checker
-_get_checker_commit_date() {
-    cd `_get_checker_code_dir` && \
-        git log -1 --format=format:'%ci'
-}
-export -f _get_checker_commit_date
-
 # add/update global attributes in a netCDF file to record the fact
 # that it has passed the checker.
-#   - compliance_checker_version (including commit has e.g. "1.1.1 (77dd26bbc8852e1aeeaf7523ab084f552d6f5fe9)")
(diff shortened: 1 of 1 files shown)


mhidas:
Note, this will fix the following errors currently seen in the pipeline logs every time a file passes the checker:
```
Jul 11 12:49:54 10-aws-syd ANMN_AM: fatal: Not a git repository (or any of the parent directories): .git
Jul 11 12:49:54 10-aws-syd ANMN_AM: fatal: Not a git repository (or any of the parent directories): .git
```

lbesnard:
@lwgordonimos this explains the email I sent you regarding "not a git repo"

lib/common/netcdf-check.sh
```diff
@@ -109,20 +95,17 @@ add_checker_signature() {
     local checker_version=`$NETCDF_CHECKER --version`
     local version_number=`echo $checker_version | egrep 'IOOS compliance checker version' | egrep -o '[0-9.]+$'`
 
-    local commit_hash=`_get_checker_commit_hash`
-    local last_updated=`_get_checker_commit_date`
+    local imos_version=$($NETCDF_CHECKER -l | grep ' imos ' | sed -r 's/.*\((.*)\)/\1/')
```
lbesnard: could do ```pip show -V cc-plugin-imos``` ?

mhidas: Yep, could do, but would still need to grep to extract just the version number from that output.

mhidas:
This is just a quick fix for now. When we migrate everything to Python, getting at these versions will be much cleaner and easier.

lib/common/netcdf-check.sh
lwgordonimos: FYI, the Chef recipe tests the versions with the following one-liners... not very pretty but serviceable until able to be queried out of a Python context.

```bash
checker_ver=$(pip show compliance-checker | awk /^Version:/'{print $2}')
plugin_ver=$(pip show cc-plugin-imos | awk /^Version:/'{print $2}')
```

PR #528, opened 2016-07-12
Add attribute compliance_checks_passed to checker signature

mhidas:
Initially I only listed the checks passed in the history attribute, but pulling out this information in the metadata harvest is much simpler if this is also included in a separate attribute.

### DIFF
--- a/lib/common/netcdf-check.sh
+++ b/lib/common/netcdf-check.sh
@@ -91,6 +91,7 @@ export -f _get_checker_code_dir
 # "$@" - checker suites passed
 add_checker_signature() {
     local file=$1; shift
+    local checks=$@
 
     local checker_version=`$NETCDF_CHECKER --version`
     local version_number=`echo $checker_version | egrep 'IOOS compliance checker version' | egrep -o '[0-9.]+$'`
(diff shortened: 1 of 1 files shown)


mhidas:
@lbesnard @ggalibert Would one of you mind merging this before 5pm?

lib/common/netcdf-check.sh
```diff
@@ -91,6 +91,7 @@ export -f _get_checker_code_dir
 # "$@" - checker suites passed
 add_checker_signature() {
     local file=$1; shift
+    local checks=$@
```
lbesnard: how do you expect checks to be equal to 'cf, imos' ? since  add_checker_signature always used to take only a file as an argument ? Are you planing to modify the generic incoming handler as well ?

mhidas: @lbesnard Look at the list of args in the comment. You need to specify the checker suites that were passed. This has always been the case. See e.g. https://github.com/aodn/data-services/blob/66fcb8b2ef265ac89357e06ebb4e192ffb9add1b/lib/common/netcdf-check.sh#L158

PR #532, opened 2016-07-14
Revert "Revert "create CARS pipeline""

lbesnard:
Reverts aodn/data-services#522


lbesnard:
@lwgordonimos I guess this is safe now

lwgordonimos:
@lbesnard this will be safe hopefully when https://github.com/aodn/chef-private/pull/1986 is merged

lwgordonimos:
@lbesnard Actually, merging this now so we can test 14-nsp after applying https://github.com/aodn/chef-private/pull/1985

Will need to revert in the unlikely event of this breaking 14.

PR #535, opened 2016-07-18
update regexp for new LJCO products

lbesnard:


lbesnard:
Another PR showing the problem of too many repos

Not to merged until tested on prod RC
## related PR
https://github.com/aodn/harvesters/pull/436
https://github.com/aodn/geoserver-config/pull/370

PR #536, opened 2016-07-20
Lock talend_trigger process for argo incoming handler

anguss00:
Prevent celery triggers from running argo talend processes if there is a process already running

### DIFF
--- a/ARGO/incoming_handler.sh
+++ b/ARGO/incoming_handler.sh
@@ -65,6 +65,17 @@ handle_additions() {
     done
 }
 
+
+# lock processing of talend trigger to prevent simultanous talend runs
+lock() {
+    exec 200>$DATA_SERVICES_TMP_DIR/argo_pid.lock
+
+    flock -n 200 \
+        && return 0 \
+        || return 1
+}
+
+
 # main
 # $1 - file to handle
 main() {
@@ -86,6 +97,11 @@ main() {
 
     log_info "Handling '$deletions_count' deletions"
     log_info "Handling '$additions_count' additions"
+
+    lock || {
+        log_info "Only one talend process can be run at once";
+        exit 1
+    }
 
     [ $additions_count -gt 0 ] && netcdf_check_added_files $manifest_file $tmp_files_added
 


ARGO/incoming_handler.sh
```diff
@@ -65,6 +65,17 @@ handle_additions() {
     done
 }
 
+
+# lock processing of talend trigger to prevent simultanous talend runs
+lock() {
+    exec 200>$DATA_SERVICES_TMP_DIR/argo_pid.lock
+
+    flock -n 200 \
```
julian1: should flock reference a file that is intended to be locked?

julian1:
Do we know what happens in the rest of the pipeline in terms of logging, temp file management, and temp file management on failure, if a process gets short-circuited by an exit() call?

julian1:
I wonder whether it might be possible to improve the scheduling so that harvesting processes don't collide. Perhaps using job queuing to address the mismatch between producer and consumer?

mhidas:
I'm sure there's a lot of room for improvement. I wonder if this PR is addressing a similar issue to what I've been seeing occasionally in the ANMN pipelines? See https://github.com/aodn/internal-discussions/issues/226

ARGO/incoming_handler.sh
julian1: Tested with @anguss00, works well.

julian1: It would be great to have a space to indicate whether it's exec called with an argument of 200, or exec redirected using the 200 filedescriptor. Eg.
```
exec 200> $DATA_SERVICES_TMP_DIR/argo_pid.lock
```

julian1:
Can we find out for sure what happens when the lock condition is triggered. Just to ensure that the handler gets called again, and with whatever resource it was supposed to handle and that it works?

If it was possible to test it by introducing a sleep or something.

julian1:
I really don't understand why we introduced a message queue - the purpose of which is to serialize the calling of process handlers. And now, we have to introduce a mutex to short-circuit execution to prevent handler collisions.

Do we know if these collisions represent normal behavior or do they only occur due to some other exceptional condition in the data-services code?

jonescc:
For info https://github.com/aodn/internal-discussions/issues/166

anguss00:
I'm going to close this PR for the reasons outlined in https://github.com/aodn/internal-discussions/issues/264. We don't need to pidlock the process IF celery is running in a correct state.

PR #539, opened 2016-07-26
ANMN acoustic pipeline to uppdate deployment metadata

mhidas:


mhidas:
Requires https://github.com/aodn/chef-private/pull/2008

mhidas:
This is ready to be merged. @ggalibert Would you mind having a quick look?

watch.d/ANMN_PA.json
```diff
@@ -0,0 +1,4 @@
+{
+  "path": [ "ANMN/Acoustic" ],
+  "execute": "ANMN/acoustic/incoming_handler.sh"
+}
```
ggalibert: Is the casing inconsistency OK with "A/acoustic" either in path or execute?

ggalibert:
Apart from that one comment above, the rest sounds good to me.

watch.d/ANMN_PA.json
mhidas: Yeah, it's ok. Those are currently the correct paths. Maybe one day I'll rename the code directory (i.e. in this repo) to make it consistent, but I don't think it's a big deal.

PR #542, opened 2016-08-03
Fix temp file cleanup in ARGO handler

lwgordonimos:


julian1:
Why is file resource creation and clean up not being localized in handled in one place? That way, the code can be written and tested once, and then left alone.

Currently would appear to contradict DRY https://en.wikipedia.org/wiki/Don't_repeat_yourself

julian1:
Merging on the basis of other discussion.

Issue #482, opened 2016-06-09
IMOS check_geospatial_lat_min_max exception


mhidas:
Running the IMOS checker plugin (https://github.com/aodn/data-services/tree/imos_checker_plugin) on [this file](http://thredds.aodn.org.au/thredds/catalog/IMOS/eMII/demos/ACORN/monthly_gridded_1h-avg-current-map_non-QC/TURQ/2012/catalog.html?dataset=IMOS/eMII/demos/ACORN/monthly_gridded_1h-avg-current-map_non-QC/TURQ/2012/IMOS_ACORN_V_20121001T000000Z_TURQ_FV00_monthly-1-hour-avg_END-20121029T180000Z_C-20121030T160000Z.nc.gz) reports the following exceptions:
```
imos.check_geospatial_lat_min_max: only length-1 arrays can be converted to Python scalars
imos.check_geospatial_lon_min_max: only length-1 arrays can be converted to Python scalars
```
This is bad.

mhidas:
The file in question has list-values geospatial_lat/lon_min/max attributes.
```
:geospatial_lat_min = -29.5386582, -29.5927878, -29.6469169, -29.7010456, -29.7551738, etc...
```
This doesn't really make sense. These attributes should be a single numeric value. However, the checker should report this as a failed check, rather than an exception.

PR #543, opened 2016-08-09
IMOS checker: fix checks on geospatial min/max attributes

mhidas:
* Allow geospatial_min/max attributes to be any numeric type, not just float
* Check for list-valued geospatial_min/max attributes (#482)
* Update unittests
* Update plugin version to 0.9.1


mhidas:
@ggalibert would you mind having a look at this one?

PR #544, opened 2016-08-18
Add additional process tree logging on harvester failure

anguss00:


AATAMS/AATAMS_sattag_dm/incoming_handler.sh
```diff
@@ -100,7 +100,8 @@ handle_additions() {
     # finally, harvest the files!
     if ! index_files_bulk $UNZIPPED_DIR $AATAMS_SATTAG_DM_BASE $valid_new_mdb_files; then
         rm -f $valid_new_mdb_files
-        file_error "Failed indexing files, aborting operation..."
+        local process_tree=`pstree -aspG $$`
```
julian1: We're going to end up as bash debugging experts.

anguss00: The world leaders in bash debugging

julian1:
My only (real) comment was whether `ps auxf` would be better, because it includes the arguments given to the process in the output, and it is normally part of a default install without needing apt-get.

But I can't see a way, to limit the output by specifying just the root process, rather than the tree of all processes.

anguss00:
What about your bash call stack trick @julian1?

PR #550, opened 2016-09-05
ANDS platform vocab - new version 1.3

lbesnard:


bpasquer:
Thanks @loz and again welcome back

PR #553, opened 2016-09-06
Change vocab xml url to content

lbesnard:
* reason : since Nat is the only one creating the new vocab version and
   pushing the xml file to content, it is more logical to use the
   content url


bpasquer:
👍

PR #556, opened 2016-09-06
Refactor IMOS checker plugin, in preparation for IMOS-1.4 checker

mhidas:
This should not change the behaviour of the checker, except for small changes in the wording of the result messages. It is just rearranging the code to make it easier to create a new checker for the IMOS-1.4 conventions with minimal additional code. The new checker will be a subclass of the IMOSChecker class, overriding only the checks that have changed since v1.3 of the conventions.

The actual 1.4 checker will be added in the next PR. Coming soon!


mhidas:
It would be great if someone could review & merge this PR this week. @bpasquer @lbesnard ?

It may look a bit daunting, but if you look at it one commit at a time it might make a bit more sense.

It does pass the unittests.

mhidas:
Note that it's save to merge this even if you're not sure it will work for all files. Deploying the new version of the checker now requires an additional manual step on Jenkins, so we can test it on RC before deploying to prod.

Also, we'll get the devs to post-review this.

lib/cc_plugin_imos/cc_plugin_imos/util.py
```diff
@@ -354,7 +358,7 @@ def check_value(name, value, operator, ds, check_type, result_name, check_priori
                                 ["Units %s should be equivalent to %s" % (retrieved_name, value)]
 
         if operator == OPERATOR_EMAIL:
-            if not is_valid_email(retrieved_value):
+            if not is_valid_email(retrieved_value)[0]:
                 passed = False
                 reasoning_out = reasoning or ["Attribute %s is not a valid email address" % \
```
lbesnard: why not using 
```is_valid_email(retrieved_value)[1]``` in the message then ?

mhidas:
Thanks @lbesnard !

PR #558, opened 2016-09-12
Automatically process burst-averaged product in ANMN pipelines

mhidas:


ANMN/common/incoming_handler.sh
```diff
@@ -21,6 +23,39 @@ regex_filter() {
 }
 
 
+# create products based on this file, if any, and push them to incoming.
+# fail with file_error if processing fails
+# $1 - input file to process
+# $2 - destination path of input file
+create_products() {
+    local file=$1; shift
+    local dest_path=$1; shift
```
bpasquer: dest_path=$2

lbesnard: no need, there is a shift

PR #559, opened 2016-09-13
add single NC file processing

bpasquer:
The incoming handler is now enabling the processing of single netcdf files. No fundamental change in the processing though.

### DIFF
--- a/SOOP/SOOP_CO2/incoming_handler.sh
+++ b/SOOP/SOOP_CO2/incoming_handler.sh
@@ -22,7 +22,7 @@ is_future_reef_map_file() {
 }
 
 # notify_recipients
-# notify uploader and backup recipient about status pof uploaded file
+# notify uploader and backup recipient about status of uploaded file
 # $1 - file name
 # $2 - message
 notify_recipients() {
@@ -43,10 +43,10 @@ handle_netcdf_file() {
 
     log_info "Handling SOOP CO2 file '$file'"
 
-    echo "" | notify_by_email $BACKUP_RECIPIENT  "Processing new underway CO2 file '$file'"
+    echo "" | notify_by_email $BACKUP_RECIPIENT "Processing new underway CO2 file '$file'"
 
     local tmp_file_with_sig
-    local checks='cf imos:1.3'
+    local checks='cf imos:1.4'
 
     if is_imos_soop_co2_file $file; then
         tmp_file_with_sig=`trigger_checkers_and_add_signature $file $BACKUP_RECIPIENT $checks`
@@ -54,8 +54,7 @@ handle_netcdf_file() {
         checks='cf'
         tmp_file_with_sig=`trigger_checkers_and_add_signature $file $BACKUP_RECIPIENT $checks`
     else
-        file_error_and_report_to_uploader $BACKUP_RECIPIENT "Not an underway CO2 file '$file'"
-        rm -f $tmp_file_with_sig $file
+        file_error_and_report_to_uploader $BACKUP_RECIPIENT "Not an underway CO2 file "`basename $file`
     fi
 
     local path
@@ -84,15 +83,15 @@ handle_zip_file() {
     if [ $? -ne 0 ]; then
         rm -f $tmp_zip_manifest
         rm -rf --preserve-root $tmp_dir
-        file_error_and_report_to_uploader  $BACKUP_RECIPIENT "Error unzipping file '$file'"
+        file_error_and_report_to_uploader  $BACKUP_RECIPIENT "Error unzipping file "`basename $file`
     fi
 
     local nc_file
-    nc_file=`grep ".*.nc" $tmp_zip_manifest | head -1`
+    nc_file=`grep ".*\.nc" $tmp_zip_manifest | head -1`
     if [ $? -ne 0 ]; then
         rm -f $tmp_zip_manifest
         rm -rf --preserve-root $tmp_dir
-        file_error_and_report_to_uploader $BACKUP_RECIPIENT "Cannot find NetCDF file in zip bundle '$file'"
+        file_error_and_report_to_uploader $BACKUP_RECIPIENT "Cannot find NetCDF file in zip bundle "`basename $file`
     fi
 
     log_info "Processing '$nc_file'"
@@ -123,9 +122,8 @@ main() {
     elif has_extension $file "nc"; then
         handle_netcdf_file $file
     else
-        file_error_and_report_to_uploader $BACKUP_RECIPIENT "Unknown file extension '$file'"
+        file_error_and_report_to_uploader $BACKUP_RECIPIENT "Unknown file extension "`basename $file`
     fi
-
 }
 
 main "$@"
--- a/SOOP/SOOP_CO2/incoming_handler.sh
+++ b/SOOP/SOOP_CO2/incoming_handler.sh
@@ -21,58 +21,59 @@ is_future_reef_map_file() {
     echo $file | egrep -q '^FutureReefMap_'
 }
 
+# notify_recipients
+# notify uploader and backup recipient about status pof uploaded file
+# $1 - file name
+# $2 - message
+notify_recipients() {
+    local file=$1; shift
+    local message=$1; shift
+    local recipient=`get_uploader_email $file`
+
+    if [ -n "$recipient" ]; then
+        echo "" | notify_by_email $recipient $message
+    fi
+    echo "" | notify_by_email $BACKUP_RECIPIENT $message
+}
+
 # handles a single netcdf file, return path in which file is stored
 # $1 - netcdf file
 handle_netcdf_file() {
     local file=$1; shift
+
     log_info "Handling SOOP CO2 file '$file'"
 
-    local recipient=`get_uploader_email $file`
-    if [ -n "$recipient" ]; then
-        echo "" | notify_by_email $recipient "Processing new underway CO2 file '$file'"
-    fi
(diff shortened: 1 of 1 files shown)


SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -21,11 +21,54 @@ is_future_reef_map_file() {
...
+# $1 - netcdf file
+handle_netcdf_file() {
+    local file=$1; shift
+    log_info "Handling SOOP CO2 file '$file'"
+
+    local recipient=`get_uploader_email $file`
+    if [ -n "$recipient" ]; then
+        echo "" | notify_by_email $recipient "Processing new underway CO2 file '$file'"
+    fi
+    echo "" | notify_by_email $BACKUP_RECIPIENT "Processing new underway CO2 file '$file'"
```
mhidas: The above 4 lines are repeated a few times in this script. Might be worth replacing defining a function to do it?

SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -21,11 +21,54 @@ is_future_reef_map_file() {
...
+
+    local tmp_file=`make_writable_copy $file`
```
mhidas: You don't need to do this. `trigger_checkers_and_add_signature` creates a writable copy and returns its name.

SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -21,11 +21,54 @@ is_future_reef_map_file() {
...
+    local tmp_file_with_sig
+    local checks='cf imos'
+
+    if is_imos_soop_co2_file $tmp_file; then
```
mhidas: why not test the original `$file` here? (and two lines below)

SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -21,11 +21,54 @@ is_future_reef_map_file() {
...
+        file_error "Not an underway CO2 file '$file'"
+        rm -f  $tmp_file_with_sig $tmp_file
+    fi
+
+    rm -f $tmp_file
+
+    tmp_file=$tmp_file_with_sig
+
+    local path
+    path=`$SCRIPTPATH/dest_path.py $tmp_file` || file_error "Cannot generate path for "`basename $tmp_file`
```
mhidas: Again, ideally it shouldn't make a difference, but why not determine the dest path from the original `$file`?

SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -21,11 +21,54 @@ is_future_reef_map_file() {
...
+
+    s3_put $tmp_file $path/`basename $file` && rm -f $file
```
mhidas: Does `$path` include the file name already?
If not, your email message below is a bit wrong...

SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -54,7 +97,6 @@ main() {
     log_info "Processing '$nc_file'"
```
mhidas: Just call your new `handle_netcdf_file` function here!
(otherwise repeat same comments as above...)

bpasquer:
@mhidas Took me a while, but here comes my updated  incoming_handler .

SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -21,58 +21,59 @@ is_future_reef_map_file() {
     echo $file | egrep -q '^FutureReefMap_'
 }
 
+# notify_recipients
+# notify uploader and backup recipient about status pof uploaded file
```
mhidas: typo "...status of..."

SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -21,58 +21,59 @@ is_future_reef_map_file() {
...
     local file=$1; shift
+
     log_info "Handling SOOP CO2 file '$file'"
 
-    local recipient=`get_uploader_email $file`
-    if [ -n "$recipient" ]; then
-        echo "" | notify_by_email $recipient "Processing new underway CO2 file '$file'"
-    fi
-    echo "" | notify_by_email $BACKUP_RECIPIENT "Processing new underway CO2 file '$file'"
+    echo "" | notify_by_email $BACKUP_RECIPIENT  "Processing new underway CO2 file '$file'"
```
mhidas: so you no longer want to notify the uploader here?

SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -21,58 +21,59 @@ is_future_reef_map_file() {
...
+    if is_imos_soop_co2_file $file; then
+        tmp_file_with_sig=`trigger_checkers_and_add_signature $file $BACKUP_RECIPIENT $checks`
+    elif is_future_reef_map_file $file; then
         checks='cf'
-        tmp_file_with_sig=`trigger_checkers_and_add_signature $tmp_file $BACKUP_RECIPIENT $checks`
+        tmp_file_with_sig=`trigger_checkers_and_add_signature $file $BACKUP_RECIPIENT $checks`
     else
-        file_error "Not an underway CO2 file '$file'"
-        rm -f  $tmp_file_with_sig $tmp_file
+        file_error_and_report_to_uploader $BACKUP_RECIPIENT "Not an underway CO2 file '$file'"
```
mhidas: Could shorten the message a bit by using `basename $file` here. The path won't mean anything to the uploader anyway.

SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -83,49 +84,26 @@ handle_zip_file() {
     if [ $? -ne 0 ]; then
         rm -f $tmp_zip_manifest
         rm -rf --preserve-root $tmp_dir
-        file_error "Error unzipping"
+        file_error_and_report_to_uploader  $BACKUP_RECIPIENT "Error unzipping file '$file'"
```
mhidas: As above, could just use `basename $file` in the error message.

SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -83,49 +84,26 @@ handle_zip_file() {
...
     fi
 
     local nc_file
     nc_file=`grep ".*.nc" $tmp_zip_manifest | head -1`
```
mhidas: If you want to match the dot before 'nc', the pattern should be `".*\.nc"` (otherwise '.' matches any character)

SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -83,49 +84,26 @@ handle_zip_file() {
...
-        rm -f $tmp_zip_manifest $tmp_nc_file_with_sig $tmp_nc_file
-        rm -rf --preserve-root $tmp_dir
-    fi
-    tmp_nc_file=$tmp_nc_file_with_sig
-
-    local path
-    path=`$SCRIPTPATH/dest_path.py $tmp_nc_file` || file_error "Cannot generate path for "`basename $tmp_nc_file`
-
-    s3_put $tmp_nc_file $path/`basename $nc_file` && rm -f $tmp_dir/$nc_file
+    local path_to_storage=`handle_netcdf_file $tmp_dir/$nc_file`
```
mhidas: Could be worth checking at this point that `$path_to_storage` is not an empty string?

SOOP/SOOP_CO2/incoming_handler.sh
bpasquer: sure

SOOP/SOOP_CO2/incoming_handler.sh
bpasquer: typo , thank you

SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -46,20 +46,20 @@ handle_netcdf_file() {
...
+    local checks='cf imos:1.4'
 
     if is_imos_soop_co2_file $file; then
         tmp_file_with_sig=`trigger_checkers_and_add_signature $file $BACKUP_RECIPIENT $checks`
     elif is_future_reef_map_file $file; then
         checks='cf'
         tmp_file_with_sig=`trigger_checkers_and_add_signature $file $BACKUP_RECIPIENT $checks`
     else
-        file_error_and_report_to_uploader $BACKUP_RECIPIENT "Not an underway CO2 file '$file'"
+        file_error_and_report_to_uploader $BACKUP_RECIPIENT "Not an underway CO2 file " `basename $file`
```
mhidas: Sorry... I think this way the file name will become a third argument to the function and not included in the message. You need to remove the space between them:
```
file_error_and_report_to_uploader $BACKUP_RECIPIENT "Not an underway CO2 file "`basename $file`
```
Or put the filename within the double quotes:
```
file_error_and_report_to_uploader $BACKUP_RECIPIENT "Not an underway CO2 file `basename $file`"
```

SOOP/SOOP_CO2/incoming_handler.sh
mhidas: Ah yes, should be.

SOOP/SOOP_CO2/incoming_handler.sh
bpasquer: of course the space. I really have issue with emptiness in general and spaces in particular !

bpasquer:
should be good now @mhidas

PR #561, opened 2016-09-20
Checker plugin for IMOS-1.4 conventions

mhidas:
This adds a new checker class for version 1.4 of the IMOS NetCDF conventions. Most of the checks are the same as for 1.3, so the common checks are now in an `IMOSBaseCheck` class.

Note that when this gets deployed, the new IMOS-1.4 checker will be the default one selected if the compliance checker is called with `--test imos`. The previous version can be explicitly selected with `--test imos:1.3`. So before I deploy the updated checker I will create another PR to update all the pipelines to use `imos:1.3` for now. In the meantime, this can be merged once it has been reviewed.


lib/cc_plugin_imos/cc_plugin_imos/tests/data/imos_test_variable.cdl
```diff
@@ -57,5 +54,5 @@ data:
 
  bad2_qc = -127 ;
 
- bad3_qc = -127 ;
+ new_qc = -127 ;
 }
```
lbesnard: isn't it worth having a cdl file per imos convention ?

mhidas: Yes, it would probably make it easier to update the tests in the future. For now I think this is ok. I can create separate cdl files in a future "clean-up" PR.

mhidas:
The PR to update the pipelines before this is deployed is here: #562

mhidas:
Thanks @lbesnard 

@ggalibert it would be great if you could post-review this a bit.

lib/cc_plugin_imos/cc_plugin_imos/tests/test_imos.py
```diff
@@ -306,7 +300,52 @@ def test_vertical_coordinate_type(self):
...
 
 
-    ### Test compliance checks
+################################################################################
+#
+# IMOS 1.3 Checker
+#
+################################################################################
+
+class TestIMOS1_3(unittest.TestCase):
```
ggalibert: Could this class inherit from TestUtils? Then you won't have to duplicate most of the code below.

mhidas: Yes, but it would also inherit all the `test_...` methods that we don't need to repeat. I'd have to create a separate base class with just the shared methods, and then `TestUtils` and `TestIMOS1_3` could both inherit from that. I can do it in a future PR.

lib/cc_plugin_imos/cc_plugin_imos/tests/test_imos.py
```diff
@@ -907,6 +907,28 @@ def test_check_quality_control_conventions_for_quality_control_variable(self):
...
+    def test_check_quality_control_global(self):
+        self.imos.setup(self.new_dataset)
+        ret_val = self.imos.check_quality_control_global(self.new_dataset)
+        self.assertEqual(len(ret_val), 4)
+        passed_var = [r.name[1] for r in ret_val if r.value]
+        good_var = ['DEPTH_quality_control', 'TEMP_quality_control']
+        self.assertEqual(len(passed_var), len(good_var) * 2)
+        self.assertEqual(set(passed_var), set(good_var))
+
+        self.imos.setup(self.test_variable_dataset)
```
ggalibert: For clarity it would be great to have variable names for test files that match the actual cdl file names:

test_variable_dataset == imos_test_variable.cdl ?

lib/cc_plugin_imos/cc_plugin_imos/tests/test_imos.py
```diff
@@ -627,19 +627,30 @@ def test_data_variable_list(self):
         self.assertEqual(self.imos._data_variables[0].name, 'data_variable')
         self.assertEqual(self.imos._data_variables[1].name, 'random_data')
 
+    def test_check_data_variable_present(self):
+        self.imos.setup(self.good_dataset)
+        ret_val = self.imos.check_data_variable_present(self.good_dataset)
+        self.assertEqual(len(ret_val), 1)
```
mhidas: Yes, that's what it *should* return, but `ret_val` is a list, so if something's gone wrong in the code, it could have additional items. All tests are here to ensure that the checks return what they should.

lib/cc_plugin_imos/cc_plugin_imos/tests/test_imos.py
```diff
@@ -723,4 +723,80 @@ def test_check_geospatial_vertical_units(self):
...
-    pass
+
+    def setUp(self):
+        '''
+        Initialize the dataset
+        '''
+        super(TestIMOS1_4, self).setUp()
+        self.imos = IMOS1_4Check()
+        # good_dataset was only all good for IMOS 1.3!
+        self.old_good_dataset = self.good_dataset
```
mhidas: Yes, at least for the "good" data sets... another task for the cleanup.

lib/cc_plugin_imos/cc_plugin_imos/tests/test_imos.py
mhidas: Indeed! Also on the TODO list.

lib/cc_plugin_imos/cc_plugin_imos/tests/test_imos.py
```diff
@@ -866,6 +865,30 @@ def test_check_coordinate_variable_no_fill_value(self):
...
+                    ('TIME', 'calendar'),
+                    ('TIME', 'valid_min'),
+                    ('TIME', 'valid_max'),
+                    ('TIME', 'units')]
+
+        ret_val = self.imos.check_time_variable(self.new_dataset)
+        self.assertEqual(len(ret_val), len(time_att))
+        passed_att = [r.name[1:] for r in ret_val if r.value]
+        self.assertEqual(len(passed_att), len(time_att))
+        self.assertEqual(set(passed_att), set(time_att))
```
mhidas: `assertEqual(len(...), len(...))` makes sure we have the right number of passed attributes.
`assertEqual(set(...), set(...))` makes sure we all the correct passed attributes, but doesn't catch a case where there are multiple return results with the same attribute name (since a `set` has only unique values). In this case it probably is a bit redundant, but I don't think it hurts.

lib/cc_plugin_imos/cc_plugin_imos/tests/test_imos.py
```diff
@@ -723,4 +723,80 @@ def test_check_geospatial_vertical_units(self):
...
+                           'license',
+                           'standard_name_vocabulary'])
+
+        ret_val = self.imos.check_mandatory_global_attributes(self.new_dataset)
+        att_passed = set([r.name[1] for r in ret_val if r.value])
+        self.assertEqual(att_passed, attributes)
+
+        ret_val = self.imos.check_mandatory_global_attributes(self.bad_dataset)
+        att_failed = set([r.name[1] for r in ret_val if not r.value])
+        self.assertEqual(att_failed, attributes)
```
mhidas: Exactly. In `new_dataset` they are all correct, and in `bad_dataset` they are all incorrect or missing.

lib/cc_plugin_imos/cc_plugin_imos/tests/test_imos.py
```diff
@@ -723,4 +723,80 @@ def test_check_geospatial_vertical_units(self):
...
+
+        ret_val = self.imos.check_mandatory_global_attributes(self.old_good_dataset)
+        att_all = set([r.name[1] for r in ret_val])
+        att_failed = set([r.name[1] for r in ret_val if not r.value])
+        self.assertEqual(att_all, attributes)
```
mhidas: `att_all` is the set of all attributes checked by `check_mandatory_global_attributes` (whether the check passed or failed).
`attributes` is the set of all attributes that *should* have been checked.

lib/cc_plugin_imos/cc_plugin_imos/imos.py
```diff
@@ -1187,3 +1179,27 @@ def check_vertical_variable_reference_datum(self, dataset):
             )
 
         return ret_val
+
+    def check_data_variables(self, dataset):
+        """
+        Check that each data variable has the required attributes:
+        - units
+        - coordinates (must be a blank-separated list of valid variable names)
```
mhidas: No. The `coordinates` attribute was not mandatory in 1.3

ggalibert: Oops sorry, I meant also ```units```.

PR #564, opened 2016-09-27
Fix IMOS-1.4 checker unittests

mhidas:
*Executive summary:* please merge this so that the unittests pass and the IMOS-1.4 checker plugin can be deployed.

*The long story:* I was pretty sure all the unittests were passing before I put up my PR for the IMOS-1.4 checker ( #561), but it turns out I missed one. The 1.3 checker had a separate method called `check_acknowledgement`, and an associated unittest. This became redundant in 1.4. However, because all the 1.3 unittests are inherited by 1.4 (as all of them should still apply, unless otherwise stated), I need to explicitly skip this test for 1.4


mhidas:
@lbesnard @ggalibert @bpasquer Could someone please review & merge this today? Thanks!

PR #569, opened 2016-10-06
Add official aws cli-client, in order to interact with S3

julian1:
@bpasquer

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,3 +17,4 @@ scipy==0.14.1
 six
 tendo==0.2.8
 click
+awscli


bpasquer:
👍

julian1:
This is untested. It's my fault if it crashes everything.

lbesnard:
> This is untested

can't this be tested in the PO box ?

lwgordonimos:
The only comment I had was to double check that the config file(s) don't clash in any way with the current s3cmd config. I strongly doubt it though, if used on a "per user basis", since the other config is stored centrally. This official one is a far better tool but the s3cmd is embedded pretty heavily in the pipeline!

@lbesnard it may not be able to be tested on PO due to the mocking that occurs. Suggest we make sure 14-nec-hob is cooked and check everything there.

julian1:
The s3cmd config sits outside the user home directories and is root permissioned, because it contains private keys.

In this case, I wanted something only for public read-only actions, eg. syncing some S3 files to a local dir to push them through the 14-nec-hob test pipeline. 

In order to avoid adding another dependency - I tried to a local read-only configuration for use with s3cmd. Unfortunately, when the S3 private key is deleted from the config file, s3cmd generates error messages that it can't find the config file (poor error reporting) at which point I gave up.

In any case the dependency is probably justifiable on the basis that this is the officially developed and supported cli tooling support for S3.

pblain:
@julian1 - will the use of this be limited to read only access?  Or putting the question another way, does it increase the risk of inadvertent deletion/corruption of objects on S3?  Keep in mind that we do not have versioning enabled on SRS or AUV data.

lwgordonimos:
The other alternative would be to use s3cmd still with the "anonymous" keys, since we publish those publicly anyway: https://help.aodn.org.au/downloading-data-from-servers/amazon-s3-servers/

@pblain In the case of either tool, as long as no-one uses 'projectofficer' user account, it will be read-only access.

julian1:
@pblain The reason I'm installing aws cli is because I couldn't manage to configure s3cmd in a limited read-only capability.  Although @lwgordonimos shows that it is possible.

pblain:
@julian1 - thanks

mhidas:
>  as long as no-one uses 'projectofficer' user account, it will be read-only access.

I usually have a terminal open where I'm logged in as 'projectofficer' because that's the only way I can move/remove/edit any files in the error directories. I can't necessarily be trusted to remember that I'm not allowed to use any s3 commands in there...

PR #572, opened 2016-10-18
IMOS-1.4 checker bug fixes

mhidas:
This fixes a couple of bugs that showed up as exceptions when I ran the checker on a wide range of IMOS files.


mhidas:
@ggalibert  Would you mind taking a look at this today?

I'll have to remove `check_fill_value` completely to allow Pete's files through, but I'd prefer to do that in a PR of its own (in case we ever decide to reverse that decision).

PR #574, opened 2016-10-24
IMOS checker: make _FillValue NaN check low priority

mhidas:
We've received somewhat convincing arguments to allow not-a-number (NaN) as a _FillValue for floating-point variables. However, we'd still like to discourage IMOS data providers from doing this, if possible.

A partial solution to this is to give the fill-value check LOW priority/weight. In practice what this means is with default settings (`--criteria=normal`), the checker will *pass* a file if *only* this check fails. However, if any MEDIUM or HIGH priority checks *also* fail, the output will include comments about any variables that have NaN as _FillValue, e.g.
```

Name                             Priority:     Score:Reasoning
--------------------------------------------------------------------------------
var                                    :3:   270/271 :  
    LONGITUDE                          :3:    15/16 :  
        long_name                      :3:     0/ 1 :  
            check_atttribute_type      :3:     0/ 1 : Variable attribute
                                                      LONGITUDE:long_name not
                                                      present
recommended                            :1:     8/14 :  
    BSIO2                              :1:     0/ 1 :  
        _FillValue                     :1:     0/ 1 : We request that _FillValue
                                                      attribures be set to a
                                                      real numeric value, not
                                                      NaN
```
It's not perfect, but perhaps slightly better than completely removing this check from the code?


mhidas:
@ggalibert @smancini 
Any thoughts?

lib/cc_plugin_imos/cc_plugin_imos/imos.py
```diff
@@ -1214,11 +1214,11 @@ def check_fill_value(self, dataset):
             if not hasattr(var, '_FillValue'):
                 continue
 
-            result = Result(BaseCheck.MEDIUM, True, ('var', name, '_FillValue'))
+            result = Result(BaseCheck.LOW, True, ('recommended', name, '_FillValue'))
             if is_numeric(type(var._FillValue)) and np.isnan(var._FillValue):
                 result.value = False
                 result.msgs = [
-                    "Attribute %s:_FillValue must have a real numeric value, not NaN" % name
+                    "We request that _FillValue attribures be set to a real numeric value, not NaN"
```
ggalibert: How about "We recommend..." instead of "request" to make clearer that this is not critical in order to pass the check.

I can see that you already modified the result display to print "recommended" instead of "var".

ggalibert:
Sounds good to me.

mhidas:
Thanks @ggalibert . I've updated the wording.

PR #576, opened 2016-10-25
Cc plugin imos pycharm review

lwgordonimos:
DO NOT MERGE, FOR REVIEW ONLY

@mhidas This is a review of the cc_plugin_imos code (as requested as a post review of #556, but kind of turned into just checking the whole codebase).

This is as a result using PyCharm as a code inspection tool (including PEP8), so should probably not be merged, but maybe can highlight a few things which the IDE is really good at picking up.

I tested it in a fresh virtual environment, with only the compliance-checker, numpy and any dependencies explicitly declared in requirements.txt. Note I haven't testing thoroughly, I just built the egg in the virtual environment and ran the checker on a random NC file, which worked as expected.

The first "code inspection" commit is the result of following the suggestions around redundant characters, methods being static, warnings about potentially unset variables etc.. The second "PEP8" commit is when following conventions around spacing, reordering imports etc. but looking now, there is some overlap on things like redundant characters (in particular '\' in the middle of lists etc.) and some things which could be considered code changes ("if not result is None:" -> "if result is not None:", for functionally equivalent but more readable code).

In addition to this, the final real complaint from the IDE is that the following methods never actually access the 'dataset' parameter that is passed to them, which may be a bug (or maybe not!):
- https://github.com/aodn/data-services/blob/cc_plugin_imos-pycharm-review/lib/cc_plugin_imos/cc_plugin_imos/imos.py#L667
- https://github.com/aodn/data-services/blob/cc_plugin_imos-pycharm-review/lib/cc_plugin_imos/cc_plugin_imos/imos.py#L1243
- https://github.com/aodn/data-services/blob/cc_plugin_imos-pycharm-review/lib/cc_plugin_imos/cc_plugin_imos/imos.py#L1272
- https://github.com/aodn/data-services/blob/cc_plugin_imos-pycharm-review/lib/cc_plugin_imos/cc_plugin_imos/imos.py#L1363

I use this tool constantly to keep the code tidy, but perhaps more usefully, picking up little things that are easy to miss without a tool specifically analysing the relationships between classes and checking things like variables being set in multiple places (i.e. being conditionally set to different things) which can cause bugs. 

In terms of the post review task, nothing else sprung out at me. Of course I can't really comment much on the data science aspects of it, but this is the approach I take towards the actual Python _code_ that I write in terms of maintaining it. If I can keep the IDE happy in terms of it's inspection, it definitely does rule out a few sources of bugs and does a pretty good job of keeping things consistent!

### DIFF
--- a/lib/cc_plugin_imos/cc_plugin_imos/imos.py
+++ b/lib/cc_plugin_imos/cc_plugin_imos/imos.py
@@ -4,25 +4,26 @@ Compliance Test Suite for the Integrated Marine Observing System
 http://www.imos.org.au/
 """
 
-import numpy as np
 import re
 from datetime import datetime
+
+import numpy as np
 from cf_units import date2num
-
-from compliance_checker.cf.util import find_coord_vars, _possibleaxis, _possibleaxisunits
 from compliance_checker.base import BaseCheck, BaseNCCheck, Result
+# noinspection PyProtectedMember,PyProtectedMember
+from compliance_checker.cf.util import find_coord_vars, _possibleaxis, _possibleaxisunits
 
-from cc_plugin_imos.util import is_monotonic
-from cc_plugin_imos.util import is_numeric
-from cc_plugin_imos.util import find_ancillary_variables
-from cc_plugin_imos.util import find_data_variables
-from cc_plugin_imos.util import find_quality_control_variables
-from cc_plugin_imos.util import find_ancillary_variables_by_variable
+from cc_plugin_imos import __version__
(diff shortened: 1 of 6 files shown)


bhasin85:
@lwgordonimos Can we close/merge this PR?

mhidas:
I think we can close this now. It was not to meant to be merged anyway. I am now using PyCharm myself, so I can easily apply these fixes the next time I update the cc plugin code. Thanks @lwgordonimos

PR #581, opened 2016-10-28
Bump matplotlib version

lwgordonimos:
@mhidas Should resolve https://github.com/aodn/internal-discussions/issues/308


mhidas:
Thanks @lwgordonimos . Let's give it a go.

PR #588, opened 2016-11-10
Modification to stop creating hourly files for any CODAR radial, vect…

ggalibert:
…or or hourly. This is now ACORN UWA responsibility.


ggalibert:
I will merge this when time is right.

PR #592, opened 2016-11-14
SRS SST - add cf and srs checks

lbesnard:


lbesnard:
@lwgordonimos 
could you please re-start the watch.d  once this is merged ? works on PObox so feel free to merge it too :)

PR #596, opened 2016-11-18
Acorn purge error dir

ggalibert:


ggalibert:
@mhidas could you please have quick look at this?

lib/common/util.sh
```diff
@@ -314,3 +314,20 @@ unzip_file() {
     )
 }
 export -f unzip_file
+
+# delete files in directory that start like the given file name
+# (useful to purge error directory when same file is eventually
+# successfully processed)
+# $1 - directory
+# $2 - file name (basename)
+delete_files_in_dir() {
```
mhidas: Would it make sense to restrict this to just deleting files in an error directory (e.g. `delete_from_error_dir`)? I.e. the first arg could just be the pipeline job name (ACORN in this case) which should be a sub-directory of `$ERROR_DIR`. Might just be a bit safer.

Actually you might not even need the first arg because there's an env variable called `$JOB_NAME`.

lib/common/util.sh
```diff
@@ -314,3 +314,20 @@ unzip_file() {
...
+    local dir="$1"; shift
+    local file="$1"; shift
+
+    log_info "Removing files in '$dir' that start with '$file'"
```
mhidas: Probably don't need to log this twice. On the other hand, if `$dir` does not exist, it would be helpful to log an error before returning.

lib/common/util.sh
```diff
@@ -314,3 +314,20 @@ unzip_file() {
...
+    if [ -d $dir ]; then
+        log_info "Removing files '$dir/${file}*'"
+        rm -f $dir/${file}*
```
mhidas: Could be worth adding a `-v` so that the names of files being deleted are also logged.

lib/common/util.sh
ggalibert: Do you think the check on error dir would be overcautious? Maybe we can do without.

ggalibert:
@mhidas I've tried to take your comments into account.

PR #604, opened 2016-11-29
New pipeline & checker for NSW OEH multi-beam products

mhidas:
* overall zip file content checks
* individual file name checks
* extract zip file to a temp directory
* work out dest path on S3
* bulk index all files in zip
* notify uploader of success or failure (copy to backup recipient)
* save report to $LOG_DIR

### DIFF
--- a/AODN/NSW-OEH/incoming_handler.sh
+++ b/AODN/NSW-OEH/incoming_handler.sh
@@ -29,7 +29,7 @@ notify_uploader() {
 # one survey from NSW-OEH
 # $1 - zip file to handle
 main() {
-    local zipfile=$1; shift
+    local zipfile="$1"; shift
 
     # set up temp unzip directory and report file
     local unzip_dir=`mktemp -d --tmpdir ${JOB_NAME}_XXXXX`
@@ -53,11 +53,14 @@ main() {
     index_files_bulk $unzip_dir $dest_path $report
     if [ $? -ne 0 ]; then
         notify_uploader $report "Contents of '`basename $zipfile`' were extracted but publishing failed"
+
+        # unindex all files previously indexed to maintain db consistency with S3
+        unindex_files_bulk $unzip_dir $dest_path $report
         file_error "Indexing failed"
     fi
 
     for file in `cat $report`; do
-        s3_put_no_index $unzip_dir/$file $dest_path/`basename $file`
+        s3_put_no_index "$unzip_dir/$file" "$dest_path/$file"
     done
 
     # email report to uploader (sort processed file names in $report)
--- a/AODN/NSW-OEH/incoming_handler.sh
+++ b/AODN/NSW-OEH/incoming_handler.sh
@@ -1,0 +1,68 @@
+#!/bin/bash
+
+export PYTHONPATH="$DATA_SERVICES_DIR/lib/python"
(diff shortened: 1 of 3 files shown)


AODN/NSW-OEH/incoming_handler.sh
```diff
@@ -0,0 +1,68 @@
...
+    if [ $? -ne 0 ]; then
+        notify_uploader $report "Failed to process '`basename $zipfile`'"
+        file_error "Zip file content failed checks"
+    fi
+    [ -n "$dest_path" ] || file_error "Could not determine destination path for file"
+
+    # index and publish extracted files
+    log_info "Publishing `cat $report | wc -l` extracted files..."
+
+    index_files_bulk $unzip_dir $dest_path $report
```
lbesnard: To be aware, by doing the bulk indexing, if one file doesn't get indexed, the previous ones will already be indexed with data being committed to the database (see the AODN NSW EOH harvester).
If this fails, s3_put_no_index wont be run. Which means that somethings will break due to having data on the database, but no physical file on S3

AODN/NSW-OEH/incoming_handler.sh
```diff
@@ -0,0 +1,68 @@
...
+
+# main - handle a zip file containing data and metadata for
+# one survey from NSW-OEH
+# $1 - zip file to handle
+main() {
+    local zipfile=$1; shift
+
+    # set up temp unzip directory and report file
+    local unzip_dir=`mktemp -d --tmpdir ${JOB_NAME}_XXXXX`
+    chmod 755 $unzip_dir  # make directory world readable
```
lbesnard: I don't really get this.

AODN/NSW-OEH/incoming_handler.sh
```diff
@@ -0,0 +1,68 @@
...
+
+    cat $report | notify_by_email $BACKUP_RECIPIENT "$subject"
+    log_info "Email report sent to '$BACKUP_RECIPIENT'"
+}
+
+# main - handle a zip file containing data and metadata for
+# one survey from NSW-OEH
+# $1 - zip file to handle
+main() {
+    local zipfile=$1; shift
```
lbesnard: probably safer to have ```"$1"```

AODN/NSW-OEH/incoming_handler.sh
mhidas: I think this was mainly for debugging purposes. Back when we were first developing the pipelines we often ended up with these temp directories left there after a pipeline process ended and we couldn't even see what was in them.

AODN/NSW-OEH/incoming_handler.sh
mhidas: Yeah, good point. In fact this could be an issue even if we indexed them one at a time, because they all get some of their metadata from the shapefile included in the zip. That's why it makes sense to do it as a bulk index.
I've added a line to unindex all files if the bulk index fails.

AODN/NSW-OEH/incoming_handler.sh
mhidas: Done

mhidas:
Thanks @lbesnard Feel free to merge it if you're happy.

PR #605, opened 2016-11-29
remove cron - FAIMMS , ANMN NRS DAR YON

lbesnard:
* since the pipeline is failing, the realtime data needs to stop


lbesnard:
@lwgordonimos

PR #609, opened 2016-12-02
Added verbose logging

bhasin85:
@anguss00 @lwgordonimos

### DIFF
--- a/lib/common/email.sh
+++ b/lib/common/email.sh
@@ -36,11 +36,15 @@ _get_uploader_ftp() {
     for log_file in `_log_files_ftp`; do
         # start stripping the path until we get the best match
         local trimmed_file=`get_relative_path_incoming $file`
+        log_info "Trimmed file '$trimmed_file'"
+
         while [[ $trimmed_file == *"/"* ]]; do # as long as string contains a slash
             local ftp_user=`test -f $log_file && sudo cat $log_file | grep ", \"/$trimmed_file\", " | grep " OK UPLOAD: " | tr -s " " | cut -d' ' -f8 | tail -1`
+            log_info "Ftp User '$ftp_user'"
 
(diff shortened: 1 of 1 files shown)


lib/common/email.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash  -v
```
lbesnard: 2 spaces

bhasin85: Oops :+1:  Thanks @lbesnard

lib/common/email.sh
```diff
@@ -72,18 +79,21 @@ export -f _get_uploader_rsync
...
 
+    log_info "Uploader '$uploader'"
     echo $uploader
 }
 export -f _get_uploader
 
 # returns the email lookup file which maps between usernames and their email
 # addresses
 _email_lookup_file() {
+    log_info "EMAIL_ALIASES '$EMAIL_ALIASES'"
```
lwgordonimos: This is not required as it is just the contents of a file, so no point logging it (just look at the file).

lib/common/email.sh
```diff
@@ -148,8 +169,13 @@ export -f send_report
 get_uploader_email() {
     local file=$1; shift
     local uploader=`_get_uploader $file`
+
+    log_info "File '$file'"
+    log_info "Uploader '$uploader'"
+
     if [ x"$uploader" != x ]; then
         local uploader_email=`_get_username_email $uploader`
+        log_info "Uploader Email '$uploader_email'"
```
lwgordonimos: It would be way cleaner to have a single log entry for these functions, since they are going to be run constantly and this will make the process.log ridiculously noisy (which it already is!) and harder to find proper log messages. One log line stating "ran this function with these inputs and the result was X".

Issue #390, opened 2016-03-21
Could not find uploader for file


mhidas:
A recent upload from one of the regular ANMN uploaders had a couple of erroneous files, but the error message was sent to me because:

```
Mar 21 11:28:56 10-aws-syd ANMN_NRS: Could not find uploader for file '/mnt/ebs/tmp/tmp.Y3UBpkI1oe/IMOS_ANMN-NRS_CDEKSTUZ_20090527T020337Z_NSNSI_FV00_Profile-SBE-19plus_C-20151214T043847Z.nc'
```

danfruehauf:
I think I actually know what the problem is. That is related to:
https://github.com/aodn/chef/commit/536cede8287dfff5748ee84bb8d8137b76416bc3

So `$INCOMING_FILE` does not refer to the uploaded file (with its full path) but to the file after it is moved into the sandboxed directory.

Possible solution: Have another variable to store `$ORIGINAL_UPLOADED_FILE` or something in this form, then change the email probing to utilize that variable.

danfruehauf:
I'm not going to take responsibility for that fix since I won't be here, but here it is:
 * aodn/chef#279
 * #391

mhidas:
The two PR's above have been merged & deployed. As far as I can tell, this is fixed.

mhidas:
This has been recurring again sporadically over the past few weeks. It's a pain because it means the uploader doesn't get notified when a file has problems and can't be published.

e.g.
```
mhidas@10-aws-syd:/mnt/ebs/log/data-services$ grep uploader process.log
Nov  9 13:50:50 10-aws-syd ANMN_NRS: Could not find uploader for file '/mnt/ebs/incoming/ANMN/NRS/IMOS_ANMN-NRS_AETVZ_20160606T000000Z_NRSNSI-ADCP_FV00_NRSNSI-ADCP-14-2016-06-Sentinel-or-Monitor-Workhorse-ADCP-63_END-20161027T022959Z_C-20161108T224245Z.nc'
Nov  9 13:50:53 10-aws-syd ANMN_NRS: Could not find uploader for file '/mnt/ebs/incoming/ANMN/NRS/IMOS_ANMN-NRS_AETVZ_20160606T000000Z_NRSNSI-ADCP_FV01_NRSNSI-ADCP-14-2016-06-Sentinel-or-Monitor-Workhorse-ADCP-63_END-20161027T022959Z_C-20161108T224245Z.nc'
[...]

PR #610, opened 2016-12-05
Prevent ANMN incoming_handler from overwrriting INCOMING_DIR variable

mhidas:
This variable is being used in the pipeline infrastructure. The ANMN/common/incoming_handler was overwriting it with a different value. This seems to have led to #390


mhidas:
@bhasin85 This should fix it.

PR #612, opened 2016-12-05
Code cleaning - use PYTHONPATH env for python imports

lbesnard:
basically mod 
```python
sys.path.insert(0, os.path.join(os.environ.get('DATA_SERVICES_DIR'), 'lib'))
from python.ship_callsign import ship_callsign_list
```

to 
```python
from ship_callsign import ship_callsign_list
```

sort imports, remove unused ones ...

NOT TESTED YET

### DIFF
--- a/SOOP/SOOP_ASF_SST_XBT/SOOP_BOM_ASF_SST.py
+++ b/SOOP/SOOP_ASF_SST_XBT/SOOP_BOM_ASF_SST.py
@@ -1,15 +1,17 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
+
+import argparse
+import fnmatch
 import os
 import re
-import sys
 import shutil
+
 from tendo import singleton
-sys.path.insert(0, os.path.join(os.environ.get('DATA_SERVICES_DIR'), 'lib'))
-from python.lftp_sync import LFTPSync
-from python.imos_logging import IMOSLogging
-import fnmatch
-import argparse
+
+from imos_logging import IMOSLogging
+from lftp_sync import LFTPSync
+
 
 def download_lftp_dat_files():
     """
--- a/SOOP/SOOP_ASF_SST_XBT/soop_xbt_pipeline/create_plot.py
+++ b/SOOP/SOOP_ASF_SST_XBT/soop_xbt_pipeline/create_plot.py
@@ -5,32 +5,19 @@
 # Institute: IMOS / eMII
 # email address: laurent.besnard@utas.edu.au
 
-import sys, os
-import tempfile
-import shutil
-
-# the next 3 lines are needed to create plots without a xserver
-MPLCONFIGDIR = tempfile.mkdtemp()
-os.environ['MPLCONFIGDIR'] = MPLCONFIGDIR
-import matplotlib
-matplotlib.use('Agg') # because no Xserver. has to be run after import before pylab
-
-import string
 import datetime
-import re
-from numpy import ma
-from netCDF4 import Dataset, num2date
-from matplotlib.pyplot import (figure, plot, xlabel, ylabel, title, show, xticks)
-import matplotlib.pyplot as plt
-import pylab
 import itertools
+import sys
+import tempfile
 
-def exit_cleanup(retval):
-    shutil.rmtree(MPLCONFIGDIR)
-    exit(retval)
+from matplotlib.pyplot import (plot, savefig, subplots, subplots_adjust, text,
+                               title, xticks)
+from netCDF4 import Dataset, num2date
+from numpy import ma
+
 
 def create_plot(netcdfFilePath):
-
+    """ create plot to be ingested in SOOP XBT NRT pipeline"""
     F                     = Dataset(netcdfFilePath, 'r', format='NETCDF4')
     cruise_id             = F.XBT_cruise_ID
     sea_water_temperature = F.variables['TEMP']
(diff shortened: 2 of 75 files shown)


SOOP/SOOP_ASF_SST_XBT/SOOP_BOM_ASF_SST.py
```diff
@@ -1,15 +1,15 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
+
+from imos_logging import IMOSLogging
+from lftp_sync import LFTPSync
+from tendo import singleton
+import argparse
+import fnmatch
```
mhidas: While you're re-ordering all these imports, it might be nice to follow [PEP 8](https://www.python.org/dev/peps/pep-0008/#id23) (though I can't say I've been doing this consistently):
> Imports should be grouped in the following order:
>   1. standard library imports
>   2. related third party imports
>   3. local application/library specific imports
>
> You should put a blank line between each group of imports.

lwgordonimos: In PyCharm, "Code -> Optimize Imports". Perfect imports, every time ;)

lbesnard:
@mhidas run the isort tool on the entire repo except the checker

SOOP/SOOP_ASF_SST_XBT/soop_xbt_pipeline/create_plot.py
```diff
@@ -5,25 +5,27 @@
...
+import matplotlib.pyplot as plt
+import pylab
+from matplotlib.pyplot import figure, plot, show, title, xlabel, xticks, ylabel
+from netCDF4 import Dataset, num2date
+from numpy import ma
 
 # the next 3 lines are needed to create plots without a xserver
 MPLCONFIGDIR = tempfile.mkdtemp()
 os.environ['MPLCONFIGDIR'] = MPLCONFIGDIR
-import matplotlib
```
mhidas: This might need a bit more thought... my guess is `MPLCONFIGDIR` needs to be set before importing `matplotllib`. So either we make an exception for the import ordering here (not ideal) or we set this environment variable elsewhere before calling this script. The second option would make sense, as I don't think it 's good practice to be setting environment variables inside scripts.

SOOP/SOOP_TMV/create_plot.py
```diff
@@ -4,28 +4,30 @@
...
+import pylab
+import pyproj
+from matplotlib.pyplot import figure, plot, show, title, xlabel, xticks, ylabel
+from netCDF4 import Dataset, num2date
+from numpy import ma
 
 # the next 3 lines are needed to create plots without a xserver
 MPLCONFIGDIR = tempfile.mkdtemp()
 os.environ['MPLCONFIGDIR'] = MPLCONFIGDIR
-import matplotlib
```
mhidas: As above...

lib/python/dataUtils.py
```diff
@@ -6,17 +6,18 @@
...
+import numpy as np
+from matplotlib.pyplot import close, figure
+from matplotlib.ticker import ScalarFormatter
+from netCDF4 import Dataset
+from scipy.io import loadmat
+
 if matplotlib.get_backend() <> 'Agg':
     matplotlib.use('Agg')
-from matplotlib.pyplot import figure, close
-from matplotlib.ticker import ScalarFormatter
```
mhidas: This needs to be another exception. `matplotlib.use()` needs to be called before importing `pyplot`  (see [matplotlib docs](http://matplotlib.org/faq/howto_faq.html#generate-images-without-having-a-window-appear)).

However, as 'pyplot' is only used in the `plotRecent()` function below, we could just move these last two imports in there. In fact it looks like this function is not used anywhere, so we could just remove the whole function and all matplotlib imports here.

SOOP/SOOP_ASF_SST_XBT/soop_xbt_pipeline/create_plot.py
lwgordonimos: If I recall correctly, this was required for the older version of matplotlib, and should now be redundant with the newer one? https://github.com/aodn/data-services/pull/581

mhidas: Ah yes, you're right! Needs to be tested though!

mhidas: By test I mean remove these lines and confirm it still works:
```python
MPLCONFIGDIR = tempfile.mkdtemp()
os.environ['MPLCONFIGDIR'] = MPLCONFIGDIR
```

lbesnard:
@mhidas did more cleaning

lbesnard:
@mhidas shall I rebase , and we merge ?

SOOP/SOOP_ASF_SST_XBT/soop_xbt_pipeline/create_plot.py
```diff
@@ -7,32 +7,17 @@
...
-import shutil
-import string
 import sys
 import tempfile
 
-import matplotlib
 import matplotlib.pyplot as plt
-import pylab
-from matplotlib.pyplot import figure, plot, show, title, xlabel, xticks, ylabel
+from matplotlib.pyplot import plot, title, xticks
```
mhidas: Since the purpose of this PR is "code cleaning", you could also make the use of pyplot functions more consistent, i.e. *either*
  `import matplotlib.pyplot as plt` and then use `plt.plot`, `plt.title` etc...
*or*
  `from matplotlib.pyplot import figure, plot, show, title, ` etc... for all pyplot functions.

Not essential, but cleaner (IMHO)

mhidas:
Other than my last comment, I'm happy to merge.

lbesnard:
@mhidas ive rebased, cleaned SOOP TMV code, and did what you suggested re pyplot as plt and matplotlib ...
it's tested and ready to go

PR #616, opened 2016-12-09
Sending NetCDF Checker report to projectofficers@emii.org.au when no …

bhasin85:
…uploader found

### DIFF
--- a/lib/common/email.sh
+++ b/lib/common/email.sh
@@ -156,6 +156,7 @@ export -f send_report
 get_uploader_email() {
     local file=$1; shift
     local uploader=`_get_uploader $file`
+    local po='projectofficers@emii.org.au'
 
     if [ x"$uploader" != x ]; then
         local uploader_email=`_get_username_email $uploader`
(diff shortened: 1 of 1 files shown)


lib/common/email.sh
```diff
@@ -165,7 +166,11 @@ get_uploader_email() {
             return 0
         fi
     fi
-    log_error "Could not find uploader for file '$file'"
+
+    local error_msg = "Could not find uploader for file '$file'"
+    log_error "'$error_msg'"
```
anguss00: Just wondering if we need to log twice here, given that we provide the error message below too?

bhasin85:
Pipeline automatically sends an email when an Uploader is not found.  So no need to replicate the functionality. As discussed with @mhidas

lib/common/email.sh
bpasquer: Not sure this functionality is needed. 1) When the uploader is not fund, the email could go to the backup recipient but not the whole group or POs are going to be inundated with emails. 2) each time POs  push a file through the pipeline we will receive an email: last week i re-pushed 150 files. I don't want to receive 150 emails...

PR #617, opened 2016-12-12
 Fix issue with ancillary files being push to S3 root /

bpasquer:
Recent changes to scripts with regards to email logging have interfered with the handling of SOOP_CO2 data. It highlighted some weekness in the script. This PR fixes the issue and hopefully improve its robustness

### DIFF
--- a/SOOP/SOOP_CO2/incoming_handler.sh
+++ b/SOOP/SOOP_CO2/incoming_handler.sh
@@ -46,24 +46,32 @@ handle_netcdf_file() {
     echo "" | notify_by_email $BACKUP_RECIPIENT "Processing new underway CO2 file "`basename $file`
 
     local tmp_file_with_sig
-    local checks='cf imos:1.4'
+    local checks
 
     if is_imos_soop_co2_file $file; then
-        tmp_file_with_sig=`trigger_checkers_and_add_signature $file $BACKUP_RECIPIENT $checks`
+        checks='cf imos:1.4'
     elif is_future_reef_map_file $file; then
         checks='cf'
-        tmp_file_with_sig=`trigger_checkers_and_add_signature $file $BACKUP_RECIPIENT $checks`
     else
         file_error_and_report_to_uploader $BACKUP_RECIPIENT "Not an underway CO2 file "`basename $file`
     fi
 
+    tmp_file_with_sig=`trigger_checkers_and_add_signature $file $BACKUP_RECIPIENT $checks`
+    if [ $? -ne 0 ]; then
+            file_error "Error in NetCDF checking"
+            rm -f  $tmp_nc_file_with_sig
+    fi
+
     local path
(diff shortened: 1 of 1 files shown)


SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -50,20 +50,31 @@ handle_netcdf_file() {
 
     if is_imos_soop_co2_file $file; then
         tmp_file_with_sig=`trigger_checkers_and_add_signature $file $BACKUP_RECIPIENT $checks`
+    if [ $? -ne 0 ]; then
```
mhidas: need to indent this line more

SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -50,20 +50,31 @@ handle_netcdf_file() {
...
+            file_error "Error in NetCDF checking"
+            rm -f  $tmp_nc_file_with_sig
+        fi
     elif is_future_reef_map_file $file; then
         checks='cf'
         tmp_file_with_sig=`trigger_checkers_and_add_signature $file $BACKUP_RECIPIENT $checks`
+    if [ $? -ne 0 ]; then
```
mhidas: indent more here too

SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -50,20 +50,31 @@ handle_netcdf_file() {
...
+            file_error "Error in NetCDF checking"
+            rm -f  $tmp_nc_file_with_sig
+        fi
     else
         file_error_and_report_to_uploader $BACKUP_RECIPIENT "Not an underway CO2 file "`basename $file`
     fi
```
mhidas: You could avoid repeating a few lines by putting the trigger_checkers after this statement...
```bash
if is_imos_soop_co2_file $file; then
    checks='cf imos:1.4'
elif is_future_reef_map_file $file; then
    checks='cf'
else
    file_error_and_report_ ...
fi

tmp_file_with_sig=`trigger_checkers_and_add_signature $file $BACKUP_RECIPIENT $checks`
if [ $? -ne 0 ]; then 
    ...
```

SOOP/SOOP_CO2/incoming_handler.sh
```diff
@@ -98,12 +109,15 @@ handle_zip_file() {
...
-        s3_put_no_index $extracted_file $path_to_storage/$file_basename
-    done
-
+    if [ -n "$path_to_storage" ];then
+        local extracted_file
+        for extracted_file in `find $tmp_dir -type f`; do
+            local file_basename=`basename $extracted_file`
+            s3_put_no_index $extracted_file $path_to_storage/$file_basename
+        done
+    else
```
mhidas: This should never happen, because `handle_netcdf_file` will have already killed the process instead of returning an empty path, but better safe than sorry I guess?

bpasquer: Problem is somehow it did return an empty path in the path as shown by the files pushed under s3/.  Better adding more safeguards then.

PR #619, opened 2016-12-14
SRS checker rename to GHRSST

lbesnard:
* renamed watch.d script to GHRSST specific
 * renamed checker from srs to ghrsst
 * increment sub version  of checker


lbesnard:
@lwgordonimos , a watch.d script has been modified in this PR. Cheers

PR #622, opened 2016-12-19
Aims improvements

lbesnard:


profile.d/util.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-export PYTHONPATH=$DATA_SERVICES_DIR/lib/python
+export PYTHONPATH=$DATA_SERVICES_DIR/lib/python:$DATA_SERVICES_DIR/lib/aims
```
mhidas: Couldn't your AIMS utils be under `lib/python`?

lbesnard: I was thinking about this too. but I think it would have to be renamed to aims_realtime_util. I wouldn't mind doing it in another PR if that's alright

mhidas: :+1:  Yep, that would be good.

PR #623, opened 2016-12-19
Use more variables to idendify mooring current products

mhidas:
This is needed to correctly classify some mooring data files from ADCP and current meter instruments, which may have only UCUR_MAG & VCUR_MAG variables (i.e. not corrected to true north)

### DIFF
--- a/lib/python/test_file_classifier.py
+++ b/lib/python/test_file_classifier.py
@@ -263,6 +263,17 @@ class TestMooringFileClassifier(unittest.TestCase):
         self.assertEqual(dest_dir, 'IMOS/ANMN/NRS/NRSROT/Velocity')
         self.assertEqual(dest_filename, filename)
 
+        filename = 'IMOS_ANMN-NRS_AETVZ_20150703T053000Z_NRSROT-ADCP_FV00_NRSROT-ADCP-1507-Workhorse-ADCP-43_END-20151023T034500Z_C-20151117T074309Z.nc'
+        testfile = os.path.join(self.tempdir, filename)
+        make_test_file(testfile, {'site_code':'NRSROT'},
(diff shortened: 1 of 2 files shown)


lib/python/test_file_classifier.py
```diff
@@ -263,6 +263,17 @@ def test_velocity(self):
         self.assertEqual(dest_dir, 'IMOS/ANMN/NRS/NRSROT/Velocity')
         self.assertEqual(dest_filename, filename)
 
+        filename = 'IMOS_ANMN-NRS_AETVZ_20150703T053000Z_NRSROT-ADCP_FV00_NRSROT-ADCP-1507-Workhorse-ADCP-43_END-20151023T034500Z_C-20151117T074309Z.nc'
+        testfile = os.path.join(self.tempdir, filename)
+        make_test_file(testfile, {'site_code':'NRSROT'},
+                       UCUR_MAG={},
+                       VCUR_MAG={},
+                       CSPD={}
+        )
```
lbesnard: this bracket could be indented with the opening one in a future commit

mhidas: Yep, could do. I think this is also ok according to [PEP 8](https://www.python.org/dev/peps/pep-0008/#id17).

lbesnard: FYI, I now use a vim plugin in my vimrc (https://github.com/tell-k/vim-autopep8)
```
Plugin 'tell-k/vim-autopep8' 
let g:autopep8_select="W293,E261,E302,F405,E225,E203,E231,E305,E502,E124,E303,E266" 
```
with a list of PEP8 I want to fix automatically. In vim I press F8, and those get fixed automatically

PR #633, opened 2017-01-12
SRS GHRSST - backlog reprocessing

lbesnard:
* script to automate the copy of SRS files already on S3 back to
   incoming/pipeline


SRS/SRS_GRIDDED/backlog_pipeline_reprocessing.py
```diff
@@ -0,0 +1,100 @@
...
+
+"""
+Script to copy and reprocess n_files_limit(default is 20) GHRSST files
+via pipeline in order not to overflow the celery queue with backlog items to
+reprocess. New files being uploaded by the facility will still be processed
+within a reasonable time frame of couple of minutes max
+
+This script should be run in a tmux/screen/background as it will take a long
+long time to run
+roughly 770000 files (10 sec/file) = 3 months of reprocessing
```
jonescc: Not sure its possible to guarantee that the server will be up continuously for 3 months - so is it possible to restart from where the script got up to if it is interrupted easily?

lbesnard: yeap, it can restart/ stop whenever

lbesnard: this actually writes a log files of files already pushed to $WIP_DIR. This log is read at each iteration, ie every 20 files

jonescc: OK.  Can see that now.

lbesnard:
@lwgordonimos FYI

SRS/SRS_GRIDDED/backlog_pipeline_reprocessing.py
```diff
@@ -0,0 +1,136 @@
...
+        if exc.errno == errno.EEXIST and os.path.isdir(path):
+            pass
+        else:
+            raise
+
+
+def list_files(fname):
+    "read content of text file into list. handle tgz file (smaller for github)"
+    if fname.endswith('.tgz'):
+        tar = tarfile.open(fname, mode="r")
```
lwgordonimos: If this only ever holds a single file, it should probably be a straight GZ rather than tar.gz. Then you could just stream the content out without the mucking around with temp dirs and cleanup?

lbesnard: thanks @lwgordonimos I didnt know that. Just done it

lbesnard:
I think this is good to go.

julian1:
@lwgordonimos 

Just out of interest, how does this feed the queue and avoid overflowing it - while also keeping throughput high? 

Do you poll how full it is and then feed new files on that basis?

If there is a queue interuption - for example 10-aws-syd went down - would that pose a risk to correct processing?

lbesnard:
@julian1 Thats the all point of this script, it's written in the script header
https://github.com/aodn/data-services/blob/fd26f1a7a497537acb54f8dea7b9ab65123caa9d/SRS/SRS_GRIDDED/backlog_pipeline_reprocessing.py#L6

and here
https://github.com/aodn/data-services/blob/fd26f1a7a497537acb54f8dea7b9ab65123caa9d/SRS/SRS_GRIDDED/backlog_pipeline_reprocessing.py#L69

The only way for me to avoid overflowing the queue, and keep the new files uploaded by the facility being processed in to check what is in incoming_dir. If there are more than n files (20 by default) in incoming, no files will be copied. The main process is in a while statement. If 10 aws goes down, the script just have to be relaunch.

julian1:
@lbesnard thanks for explaining and good job!

lwgordonimos:
Late to the party, but Loz performed all the magic, wrote the script and explained it nicely! This was the way we could think of to allow regular incoming files to interleave with the reprocessing ones, in the absence of an easy way to enforce queue priorities (without mucking with separate queues).

If 10-aws-syd goes down, ALL incoming processing would be offline anyway, reprocessing files or not...

julian1:
Just out of interest - is there a way to use the rabbit mq queue api to query the necessary information? As an alternative to going the very low-level way and counting the number of files yet to process in the incoming directories? 

I ask, because I think I see same kind of thing in another application of queue to job workflow. Ideally, I the queue api should be able to tell us this information - but perhaps that's not correct?

lwgordonimos:
Yes, but would require the 'pika' library to be installed. These snippets retrieve the queue count:

```python
import pika

QUEUE= 'SRS_GHRSST_GRIDDED'

connection = pika.BlockingConnection()
channel = connection.channel()
q_len = channel.queue_declare(queue=QUEUE,
                              durable=True,
                              exclusive=False,
                              auto_delete=False
                             ).method.message_count

print "Num messages in {0} queue: {1}".format(QUEUE, q_len)
```
^ If running in a loop, would need to make sure that a) holding a connection didn't interfere with normal operaiton, or b) a new connection each iteration

```bash
$ sudo rabbitmqctl list_queues | grep SRS_GHRSST_GRIDDED
```
^ Requires root, so not overly useful...

lbesnard:
that's interesting. Thanks, will look at it

lwgordonimos:
I ran it as my normal user out of a virtualenv with pika installed, but we can install it to the system Python if need be.

julian1:
I've forgotten how this is all supposed to work - but if the queue is stopped (for whatever reason) will files simply accumulate in incoming? That might risk consuming all space, and then losing files. Loz's initial file/dir testing might be the more robust approach.

lbesnard:
this is right @julian1 . I'm wondering how the queue works as well. Sometimes on the PO box, it stops randomely even with files in incoming dir. A restart of celery is needed

julian1:
I think some of the difference is that with a conventional queue implementation, the item would be put on the queue as the first step, and only then processed to the filesystem by a queue reader/worker. While here (out of necessity), we do the reverse.

lwgordonimos:
Even in that example, it implies knowing what the "item" to be processed is via some other state external to the queue (e.g. an application with request coming in through an API). In this case, the external state happens to be the presence of the file in the incoming area of the filesystem.

The inotify daemon triggers on file creation and adds the queue item before any processing is done, so it is still the first step in this case as well. The creation *event* results in the queue item, not any actual processing of the file.

PR #636, opened 2017-01-16
S3 utils

lbesnard:
* add function to list s3 object version/info
 * add function to download s3 object specific version


bpasquer:
works well @lbesnard 
you could add a reference to the doc https://github.com/aodn/chef/blob/master/doc/README.data-tools.md

lbesnard:
thanks @bpasquer , done it !

PR #642, opened 2017-01-19
Fixes for SOOP_CO2 handler: hanging on Wakmatha path validation and s…

lwgordonimos:
…3_put stdout causing non-NC files in ZIPs to not be uploaded

@bpasquer This should fix those couple of issues!


bpasquer:
Simple yet efficient. Thanks heaps @lwgordonimos

PR #643, opened 2017-01-23
Argo - un-indexed files from DB when indexed fails on bulk operation

lbesnard:


lbesnard:
This is an improvement, However I'm not sure this would be good to merge right now as it would be harder to find indexed files not properly harvested.

see issue https://github.com/aodn/content/issues/258

lbesnard:
@bpasquer Do you think you could have a look at this please ?

bpasquer:
I think i can @lbesnard

PR #651, opened 2017-02-10
SOOP_TMV major upgrade

bpasquer:
- renamed destPath.py and function to comply with python naming
conventions
- add new processing of NRT files
- add new processing of FV00,FV01 files
- checks upgraded to imos:1.4


SOOP/SOOP_TMV/dest_path.py
```diff
@@ -0,0 +1,157 @@
...
+# -*- coding: utf-8 -*-
+
+import os
+import sys
+
+from ship_callsign import ship_callsign_list
+
+
+def ships():
+    return ship_callsign_list()
```
mhidas: You can achieve the same thing with:
```python
from ship_callsign import ship_callsign_list as ships
```

SOOP/SOOP_TMV/dest_path.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def get_platform_code(file, mode):
+    """
+    # get code from netcdf file name
+    """
+    return get_file_parts(file, mode)[4]
+
+
+def get_file_parts(file, mode):
```
mhidas: Looks like `mode` is not used in this function, so you don't need it as an argument. Same for `get_platform_code` above.

SOOP/SOOP_TMV/dest_path.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    # extract components of netcdf file name
+    """
+    file_parts = os.path.basename(file).split("_")
+    # check validity of filename
+    assert len(file_parts) >= 4, 'Filename should have at least 5 components'
```
mhidas: Should that be `len(file_parts) >= 5` then?

PR #653, opened 2017-02-14
SOOP-BA upgrade

bpasquer:
- add processing of single NC file
- fix notifications to uploader
- python code follow naming conventions
- code ready for checking CF and IMOS netcdf compliance


SOOP/SOOP_BA/helper.py
```diff
@@ -1,9 +1,12 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
+<<<<<<< HEAD
```
lbesnard: please remove diff

SOOP/SOOP_BA/helper.py
```diff
@@ -12,80 +15,89 @@
 def ships():
     return ship_callsign_list()
```
lbesnard: please do as Marty's comment in prior PR

PR #654, opened 2017-02-14
 Fix path to realtime 10sec folder plus minor changes

bpasquer:
made some changes and improvements after @mhidas comments


SOOP/SOOP_TMV/incoming_handler.sh
```diff
@@ -72,6 +72,9 @@ handle_zip_file() {
     path_to_archive=`$DATA_SERVICES_DIR/SOOP/SOOP_TMV/dest_path.py set_destination_path $zip_file 'archive' 'NRT'` 
     
     move_to_archive $zip_file IMOS/$path_to_archive
+
+    # cleaning
+    rm -rf --preserve-root $tmp_dir
```
mhidas: You could replace this with a `trap` statement just after creating the temp directory (see e.g. [here](https://github.com/aodn/data-services/blob/91fb4d49c39abaef852781571c35a7ae19389065/ARGO/incoming_handler.sh#L82)), so it will be deleted no matter how the handler exits.

PR #658, opened 2017-02-15
    SOOP-BA upgrade

bpasquer:
- add processing of single NC file
- fix notifications to uploader
- python code follow naming conventions
- code ready for checking CF and IMOS netcdf compliance
- add trap to clean tmp files


lbesnard:
checked with @bpasquer

PR #670, opened 2017-03-09
create_plote script for SOOP_TMV FV02 files needs to catter for new F…

ggalibert:
…V02 files produced without ancillary QC variables

### DIFF
--- a/SOOP/SOOP_TMV/create_plot.py
+++ b/SOOP/SOOP_TMV/create_plot.py
@@ -144,24 +144,25 @@ def create_plot_transect(netcdf_file_path, tmp_dir, param1, param2, transect):
     p2 = netcdf_file_obj.variables[param2]
     # mask data where lat or lon sert to fillvalue
 
-    # plot only data with QC flag =1, 2 (good data, probably good data)
-    no_qc = 0
-    good_flag = [1, 2]
-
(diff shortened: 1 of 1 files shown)


SOOP/SOOP_TMV/create_plot.py
```diff
@@ -144,24 +144,25 @@ def create_plot_transect(netcdf_file_path, tmp_dir, param1, param2, transect):
...
-    good_p2_idx = (netcdf_file_obj.variables[param2 + '_quality_control'][:] != no_qc ) & \
-        (netcdf_file_obj.variables[param2 + '_quality_control'][:] <= good_flag[1])
-
     p1_values = p1[:]
     p2_values = p2[:]
 
-    # replace unwanted "bad" values with the Fillvalue
-    p1_values[~good_p1_idx] = p1._FillValue
-    p2_values[~good_p2_idx] = p2._FillValue
+    if [param1 + '_quality_control'] in netcdf_file_obj.variables.keys():
```
bpasquer: it is best to follow python scripting style : the indentation  should be of multiple of four

ggalibert: Thanks, I've just fixed this.

PR #686, opened 2017-04-24
ABOS_ASFS: turn off compliance checks

mhidas:


mhidas:
The real-time files processed by this pipeline will not be made compliant until later this year. We'll turn the checks back on after that.

Issue #684, opened 2017-04-12
IMOS geospatial_vertical check needs to account for positive attribute


mhidas:
The IMOS checker plugin checks that global attributes `geospatial_vertical_min` and `geospatial_vertical_max` match the range of values in vertical coordinate variables. However, both the global attributes and the variables can have a corresponding `positive` attribute, indicating whether it is measured as positive up or down (i.e. depth or height). This obviously affects the numerical values of the min and max, but is currently not accounted for in the checker code.

PR #689, opened 2017-05-02
IMOS checker: two fixes in check_geospatial_vertical_min_max

mhidas:
This PR fixes two issues in the same check method:
* an exception that was happening for files with bad vertical coordinate variables; and
* taking the `positive` attributes into account when comparing the vertical variable ranges to the `geospatial_vertical_min/max` global attributes (#684).


mhidas:
@ggalibert would you mind taking a look at this?

In particular I'd like to know what you think of my solution in the second commit (https://github.com/aodn/data-services/pull/689/commits/9eb19effd490a445ab514f68d27afae05c6bc1f8). The idea is to only switch the sign of the coordinate values if we know for sure that the variable uses a different direction than the global attributes. Otherwise, assume that they are the same (rather than e.g. assume positive = "down").

lib/cc_plugin_imos/cc_plugin_imos/imos.py
```diff
@@ -375,8 +376,15 @@ def check_geospatial_vertical_min_max(self, dataset):
         for var in vert_vars:
             if np.isnan(var.__array__()).all() or (hasattr(var[:], 'mask') and var[:].mask.all()):
                 continue   # no valid values
-            obs_mins[var.name] = np.nanmin(var.__array__())
-            obs_maxs[var.name] = np.nanmax(var.__array__())
+
+            # account for differences in positive orientation, if known for both the variable and the global attributes
+            positive = 1
```
mhidas: Ok. Renamed to `vertical_positive_sign`

mhidas:
Yes, it's true that a file with any of the `positive` attributes missing is not IMOS compliant, but that is already checked independently by other check methods ([check_geospatial_vertical_positive](https://github.com/aodn/data-services/blob/master/lib/cc_plugin_imos/cc_plugin_imos/imos.py#L1132) and [check_vertical_variable](https://github.com/aodn/data-services/blob/master/lib/cc_plugin_imos/cc_plugin_imos/imos.py#L668)). My thought was just to let this check do something simple and sensible anyway.

> data min == global max and data max == global min

I *think* I know what you mean, but I don't think it's worth doing this, nor is it fully correct. These global attributes are defined by [ACDD](http://wiki.esipfed.org/index.php/Attribute_Convention_for_Data_Discovery) as 
> **geospatial_vertical_min** Describes the *numerically* smaller vertical limit; 
> **geospatial_vertical_max** Describes the *numerically* larger vertical limit;

I interpret that to mean that global *min* should always be equal to the data *min*, except for a possible factor of -1 if the positive directions are different.

PR #693, opened 2017-05-25
SOOP XBT DM - improvements

lbesnard:
* improve logging when a line is not defined
 * add error handling when a line  contains a forward slash / in the xbt
   line as this ends in the netcdf filename
 * improve output_folder handling so ~/ and relative directories can be
   used


lbesnard:
@smancini 
Branch can be checked by Bec Cowley.

Various lines need to be added in the config file. 

Warning in the BOM2016 data, it seems like some XBT lines were renamed, for example PX1 to PX01, or also the order. This can be done in the config file as well

New ACT codes can be added in the config file , such AS 
PE
SA
NU
PE
PS
CT

PR #707, opened 2017-06-05
All references to SST replace by sea water temperature -TEMP

bpasquer:


AODN/MHL/process_MHLsst_from_txt.py
```diff
@@ -248,6 +248,6 @@ def parse_nc_attribute(input_netcdf_file_path, output_nc_obj):
 
     """
 
-    for txtfile in glob.glob('/vagrant/tmp/MHL/TXTFILES/*.TXT'):
+    for txtfile in glob.glob('/vagrant/tmp/MHL/TXTFILES/*_new.txt'):
```
mhidas: It would be cleaner to parse the input path from the command line, rather than hard-coding it here. It only takes a couple of lines with the `argparse` module. See e.g. [here](https://github.com/aodn/data-services/blob/master/AODN/NSW-OEH/process_zip.py#L325)

bpasquer:
I've implemented your suggestion  @mhidas Thanks

AODN/MHL/process_MHLsst_from_txt.py
```diff
@@ -247,7 +248,14 @@ def parse_nc_attribute(input_netcdf_file_path, output_nc_obj):
...
+    parser = argparse.ArgumentParser()
 
-    for txtfile in glob.glob('/vagrant/tmp/MHL/TXTFILES/*_new.txt'):
+    parser.add_argument('dir_path', help="Full path to input text file directory")
+    parser.add_argument('txt_file', help="input text file")
+
+    args = parser.parse_args()
+    dir_path = args.dir_path
+    txt_file = args.txt_file
+    for txtfile in glob.glob(os.path.join(dir_path,txt_file)):
```
mhidas: That works, though if you're just going to join `dir_path` and `txt_file` together anyway, why not just have a single input arg for the full path to the file (including glob pattern)?

bpasquer:
@mhidas I've change it as you suggested  and rebased everything.

PR #716, opened 2017-06-21
 ANFOG DM : add error handling when invalid path is generated

bpasquer:


bpasquer:
fix to https://github.com/aodn/content/issues/285#

PR #717, opened 2017-06-28
CARS - handle monthly split product

lbesnard:


CSIRO/CARS/incoming_handler.sh
```diff
@@ -15,7 +15,7 @@ regex_filter() {
 # $1 - file to handle
 main() {
     local file=$1; shift
-    local regex="^CARS[[:digit:]]{4}_(World_monthly|Australia_weekly)\\.nc$"
+    local regex="^CARS[[:digit:]]{4}_(World_monthly|Australia_weekly).*\\.nc$"
```
ggalibert: @smancini told me we would keep the original monthly file on S3/THREDDS but it won't be harvested. I don't think the regex above would achieve that. Also, only the monthly product is targeted for this.

I would go for something like:
```BASH
local regex="^CARS2009_(World_monthly_[[:digit:]]{2}|Australia_weekly)\\.nc$"
```
If that makes sense. By the way, you can note that I removed the possibility to harvest a file named CARSxxxx, this was not necessary.

File names are as follow:
```
CARS2009_World_monthly_01.nc
CARS2009_World_monthly_02.nc
CARS2009_World_monthly_03.nc
CARS2009_World_monthly_04.nc
CARS2009_World_monthly_05.nc
CARS2009_World_monthly_06.nc
CARS2009_World_monthly_07.nc
CARS2009_World_monthly_08.nc
CARS2009_World_monthly_09.nc
CARS2009_World_monthly_10.nc
CARS2009_World_monthly_11.nc
CARS2009_World_monthly_12.nc
```

lbesnard: I actually checked on my pobox and it worked. Both original and new monthly files can be harvested (not at the same time though because there is a constraint on the unique time value per collection set up in the harvester). 

The thing is we still need to do a po_s3_del to unharvest the original file from the generic timestep schema and then remove the file from S3. Afterwards, we will need to do a async-upload (to bypass the pipeline) of the previous format so it can be back on S3. 

The po_s3_del command relies on the regexp set up in the chef(and chef-private) repo.   
it would be good to have a similar regex as in https://github.com/aodn/chef/pull/536/files. 
if we change the regepx in the chef repo(s) as suggested in the comments you've just added, we wouldn't be able to call po_s3_del.

I'm easy, just decide the order you want things to be done.

ggalibert: Ok, then how about:
1. We delete and unharvest the original monthly file,
2. we change the regex everywhere needed,
3. we upload both original and splitted monthly files and the splitted files only should be harvested.

lbesnard: happy with that. I ll get this sorted

lbesnard:
@ggalibert this is ready to be merged. Same for chef(s) PRs. I'll test on RC first

PR #718, opened 2017-06-28
ABOS_ASFS: create correct path for "AZFP" echo-sounder files

mhidas:
... and do a bit of code cleaning up.

### DIFF
--- a/ABOS/ASFS/dest_path.py
+++ b/ABOS/ASFS/dest_path.py
@@ -11,7 +11,6 @@ Output:
 
 Assume: (will be checked by handler)
  * File is netCDF
- * File was produced by the Toolbox
  * Site code has been validated in checker (if exists)
 """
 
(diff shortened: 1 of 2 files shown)


mhidas:
@lbesnard Would you mind taking a quick look? Best to view the two commits separately, as the first is only code clean-up, while the second is new functionality. Thanks.

mhidas:
Never mind, I'll just merge it myself. :sunglasses:

mhidas:
@pblain Here's an example of a "content" change that you wanted to review. I have tested it in RC and it works as expected. In this case I am not going to wait for a review because
1) I won't be here for the next two reviews; and
2) The facility staff have already been waiting for a month to have these files published in the right place. I am not going to make them wait another month.

So I will deploy this to prod, but feel free to add this to the post-review list.

By the way, even if I don't deliberately deploy it to prod now, this change will end up there the next time anyone runs the data_services_prod_merge job on Jenkins, because it's already on master.

PR #719, opened 2017-06-29
mailrc file is broken and unnecessary on Xenial

lwgordonimos:
Fix for https://github.com/aodn/issues/issues/73


mhidas:
Thanks @lwgordonimos Seems to work ok on 4-nec-hob.

PR #720, opened 2017-07-05
 New piepline for AATAMS acoustic :

bpasquer:
- incoming handler
- csv file checker
- watch.d

### DIFF
--- a/AATAMS/AATAMS_acoustic/check_csv.py
+++ b/AATAMS/AATAMS_acoustic/check_csv.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python
+"""
+Check that csv file has required fields
+"""
+import os
+import pandas
+import sys
+
+EXPECTED_HEADERS = {'transmitter_id', 'installation_name', 'station_name', 'receiver_name', 'detection_timestamp',
+                    'longitude', 'latitude', 'sensor_value', 'sensor_unit', 'FDA_QC', 'Velocity_QC', 'Distance_QC',
+                    'DetectionDistribution_QC', 'DistanceRelease_QC', 'ReleaseDate_QC', 'ReleaseLocation_QC',
+                    'Detection_QC'}
+
+
+def main(csvfile):
+    # CSV file to extract array of formatted data
+    data = pandas.read_csv(csvfile, delimiter=';', header=0)
+    actual_headers = set(data.columns.values)
+
+    # compare expected with actual headers, do nothing if they are equal
+    if actual_headers != EXPECTED_HEADERS:
+        only_in_actual = list(actual_headers.difference(EXPECTED_HEADERS))
+        only_in_expected = list(EXPECTED_HEADERS.difference(actual_headers))
+
+        error_message = [
+            "Columns in file '{csvfile}' don't match expected columns".format(csvfile=csvfile)]
+        # print report on what the differences were between the
+        # expected headers and the actual headers
+        if only_in_actual:
+            error_message.append("Unexpected columns: {only_in_actual}".format(
+                only_in_actual=only_in_actual))
+        if only_in_expected:
+            error_message.append("Missing columns: {only_in_expected}".format(
+                only_in_expected=only_in_expected))
+
+        sys.exit(os.linesep.join(error_message))
+
+
+if __name__ == '__main__':
+    try:
+        input_file = sys.argv[1]
+    except IndexError:
+        sys.exit("usage: {script} CSVFILE".format(script=sys.argv[0]))
+    main(input_file)
--- a/AATAMS/AATAMS_acoustic/incoming_handler.sh
+++ b/AATAMS/AATAMS_acoustic/incoming_handler.sh
@@ -1,0 +1,89 @@
+#!/bin/bash
+# AATAMS_ACOUSTIC pipeline :
+# input : ZIP file
+#
+# Requirements: the Zip archive must contain a TagMetadata.txt file 
+# Datafiles must comply with format( required number of data field)
+#
+# B. Pasquer July2017
+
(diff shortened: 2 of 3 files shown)


bpasquer:
Would be good if we could have a look at this one together tomorrow @lwgordonimos . It's doing its job fine on the PO box but doing something odd at the end

AATAMS/AATAMS_acoustic/check_csv.py
```diff
@@ -0,0 +1,22 @@
+#! /usr/bin/env python
+"""
+Simple check (number of columns) of csv file format
```
mhidas: Do you just want to check the number of columns, or should the column headers also match the required names?

bpasquer: Since i list the column names i'm expecting i should probably check for their presenceI'll do the  modification
thanks @mhidas

bpasquer:
I've updated the checking script @mhidas

AATAMS/AATAMS_acoustic/check_csv.py
```diff
@@ -0,0 +1,21 @@
...
+
+csvfile = sys.argv[1]
+header_col = ['transmitter_id', 'installation_name', 'station_name', 'receiver_name', 'detection_timestamp', 'longitude', 'latitude',
+              'sensor_value', 'sensor_unit', 'FDA_QC', 'Velocity_QC', 'Distance_QC', 'DetectionDistribution_QC', 'DistanceRelease_QC',
+              'ReleaseDate_QC', 'ReleaseLocation_QC', 'Detection_QC']
+
+# CSV file to extract array of formatted data
+data = pandas.read_csv(csvfile, delimiter=';', header=0)
+
+if set(data.columns.values).issubset(header_col) == False:
```
mhidas: According to the [docs](https://docs.python.org/2/library/stdtypes.html#set.issubset), this is testing that every column in the data is in `header_col`, not the other way around. From the error message below, it looks like you actually want
```python
if not set(header_col).issubset(data.columns.values):
```
i.e. to check that every that every element of `header_col` is present in the data!

Or, the columns in the data should be exactly the ones listed in `header_col`, with no additional ones, then simply
```python
if not set(data.columns.values) == set(header_col):
```

bpasquer: I am testing that the statement  -every columns of the data is in the header_col - is false . I agree it is overcomplicated but it is right.
Anyway i'll adopt your suggested command

bpasquer:
@xhoenner please feel free to test and deploy once this is merged.

AATAMS/AATAMS_acoustic/check_csv.py
```diff
@@ -0,0 +1,21 @@
+#! /usr/bin/env python
```
lwgordonimos: Feel free to take with a grain of salt... I'd probably expand on it to take more advantage of the set operations to report on what the differences actually were to save time digging and manually comparing, e.g.:

```python
#!/usr/bin/env python
"""
Check that csv file has required fields 
"""
import os
import pandas
import sys

EXPECTED_HEADERS = {'transmitter_id', 'installation_name', 'station_name', 'receiver_name', 'detection_timestamp',
                    'longitude', 'latitude', 'sensor_value', 'sensor_unit', 'FDA_QC', 'Velocity_QC', 'Distance_QC',
                    'DetectionDistribution_QC', 'DistanceRelease_QC', 'ReleaseDate_QC', 'ReleaseLocation_QC',
                    'Detection_QC'}


def main(csvfile):
    # CSV file to extract array of formatted data
    data = pandas.read_csv(csvfile, delimiter=';', header=0)
    actual_headers = set(data.columns.values)

    # compare expected with actual headers, do nothing if they are equal
    if actual_headers != EXPECTED_HEADERS:
        only_in_actual = list(actual_headers.difference(EXPECTED_HEADERS))
        only_in_expected = list(EXPECTED_HEADERS.difference(actual_headers))

        error_message = ["Columns in file '{csvfile}' don't match expected columns".format(csvfile=csvfile)]
        # print report on what the differences were between the expected headers and the actual headers
        if only_in_actual:
            error_message.append("Unexpected columns: {only_in_actual}".format(only_in_actual=only_in_actual))
        if only_in_expected:
            error_message.append("Missing columns: {only_in_expected}".format(only_in_expected=only_in_expected))

        sys.exit(os.linesep.join(error_message))


if __name__ == '__main__':
    try:
        input_file = sys.argv[1]
    except IndexError:
        sys.exit("usage: {script} CSVFILE".format(script=sys.argv[0]))
    main(input_file)
```
Just exits silently with success if they match, otherwise prints handy report on what the problem is, e.g.

```bash
/home/lwgordon/check_csv/bin/python /home/lwgordon/check_csv.py test.csv
Columns in file 'test.csv' don't match expected columns
Unexpected columns: ['dsadsa.2', 'dsadsad', 'dsadsa.1', 'dsadsa']
Missing columns: ['installation_name', 'transmitter_id', 'ReleaseLocation_QC', 'sensor_unit', 'ReleaseDate_QC', 'detection_timestamp', 'longitude', 'station_name', 'Detection_QC', 'sensor_value', 'DetectionDistribution_QC', 'Velocity_QC', 'latitude', 'receiver_name', 'FDA_QC', 'DistanceRelease_QC', 'Distance_QC']

Process finished with exit code 1
```

bpasquer: This is very close to perfection!( yes only very close, cause who knows what perfection is...)
Thanks for the time you spent on commenting on the script. This is a good reminder of the good programming practices i should embrace...

bpasquer:
I've adopted you python script @lwgordonimos .
I have aslo made the change to remove the zip file after processing .No more error message at the end.

PR #727, opened 2017-07-27
Tag metadta file now copied to WIP directory and to S3

bpasquer:


bpasquer:
@xhoenner Note that i don't remove the Tagmetadata file from the wip dir after use. Let me know if i should add a bit of cleaning up in the script

PR #730, opened 2017-08-02
Simplified regex to allow 10 and 50 m resolution files

bpasquer:


bpasquer:
Maybe you could have a look at this one pls @lwgordonimos

PR #736, opened 2017-09-06
XBT DM - use ands vocab for processing

lbesnard:
* create new lib routines to read ands XBT line vocab
 * use new lib with the XBT python code processing


lbesnard:
@smancini FYI

PR #739, opened 2017-09-18
AUV - divide by the number of threads used to process a AUV campaign

lbesnard:
@ggalibert

### DIFF
--- a/AUV/auv_viewer_processing/auv_processing.py
+++ b/AUV/auv_viewer_processing/auv_processing.py
@@ -210,7 +210,11 @@ def generate_geotiff_thumbnails_dive(geotiff_dive_list, thumbnail_dir_path):
 
     partial_job = partial(_generate_geotiff_thumbnail, thumbnail_dir_path)
     n_cores     = cpu_count()
-    pool        = Pool(n_cores)
+    if n_cores > 1:
+        pool = Pool(2)  # only use 2 cores to let other processes run smoothly
+    else:
+        pool = Pool(1)
+
     pool.map(partial_job, geotiff_dive_list)
     pool.close()
     pool.join()


AUV/auv_viewer_processing/auv_processing.py
```diff
@@ -210,7 +210,7 @@ def generate_geotiff_thumbnails_dive(geotiff_dive_list, thumbnail_dir_path):
 
     partial_job = partial(_generate_geotiff_thumbnail, thumbnail_dir_path)
     n_cores     = cpu_count()
-    pool        = Pool(n_cores)
+    pool        = Pool(n_cores / 2)  # only use half of cores to let other processes run smoothly
```
ggalibert: @lwgordonimos would that be sufficient or do you recommend a different approach? My understanding is this process is run once every 6months and doesn't need to be going as fast as possible.

lwgordonimos: It should definitely be lowered, but as to whether it is sufficient, it's impossible to say without having measured the difference in load. 

n_cores will be equal to 8 on production, so this would reduce the pool to 4. It may be that 4 is still quite aggressive, and it's also very likely that too many processes would just end up slowing each other down (as well as the other processes).

@lbesnard do you have an idea of the rough difference in time between 1/2/4/8 cores?

lbesnard: For this campaign, the processing time was 6 hours (8 cores). so roughly 48 hours of core work hours.
n= 1 -> 48 hours
n = 2 -> 24 hours
n = 4 -> 12 hrs
n = 8 -> 6 hrs

The process takes some time because of the creation of image thumbnails. There is a lot of I/O.

lbesnard: But if we only use 1 core, how to be sure ACORN won't use the same core ?

lwgordonimos: The IO was why I was wondering if it was linear time like that.

Processes are allocated to cores by the OS, so it won't use the same core, since it will schedule the processes however it determines as optimal. We just need to make sure the overall load isn't high enough to hog a high proportion of the total capacity.

That said, it may be that the IO load is more of a problem, in which case the cores might be nearly irrelevant to the load, and the *number* of processes doing IO on the same volume is a real problem. In this case it might be better to decouple the pool size from the core count completely and just set a fixed value relevant to the IO capacity of the volume.

I personally think for this to run on 10-aws-syd we should maybe chop it down to 2, since 10-aws-syd runs so many things. Ideally, this kind of work would be done on a separate server entirely... could it be run on a Nectar server and the products be uploaded to prod?

lbesnard: well it could be done on nectar, but only the aws10 machine is allowed to rsync the AUV data from the AUV facility. Secondly if there is no pipeline set up on the nectar machine, it means that all the process which is currently automated would become a manual process ... -> making mistakes. I'm definitely not keen

ggalibert: Would using 1 or 2 cores be acceptable for you? Everyone else uses 1 core max on this machine.

lwgordonimos: Sorry, I was under the impression this was a manually triggered task anyway. I definitely don't want to de-automate anything!

I think @ggalibert is correct. In the absence of another server to do this kind of processing, if it has to run on 10-aws-syd, we just need to limit the resource usage down to 1 or 2 max.

lbesnard: the task is triggered manually, but it's basically just launching one command which then does a lot of automated tasks. Will change the PR to use 2 cores. I have another campaign to process

lbesnard:
ready to be merged

PR #741, opened 2017-09-19
Add watch/handler files for ACORN_DM processing

lwgordonimos:
WIP: do not merge until complete!


ggalibert:
At terms, when this is deployed and working, we will want to update the ACORN incoming handler's regex so that it only deals with FV00 (RT) files.

ggalibert:
Since there's only a change in the regex for each incoming handler, there might be a more elegant way of doing this rather that duplicating the incoming handler's code in both ACORN and ACORN_DM BASH directories.

lwgordonimos:
@ggalibert looks reasonable to me. When you're around next, we can merge and deploy to RC to complete testing and then arrange for production and notifying the facility.

ggalibert:
A bit weird this file 'File not handled'... Not sure where that comes from.

ggalibert:
This kind of error already exists in production and is not limited to ACORN.

PR #748, opened 2017-10-03
Update Deakin bathy regex to allow .tif files to be archived.

ggalibert:


ggalibert:
@bpasquer or @lbesnard could you please have a look?

bpasquer:
What do you mean by "archive" the .tif files? Storing them on S3 alongside the NetCDF @ggalibert or do you want to push them to the archives?

ggalibert:
I want to have it accessible on S3 alongside the NC files.

PR #750, opened 2017-10-04
Update ANMN incoming handler to handle bursting instruments other tha…

ggalibert:
…n WQM and NXIC (for example any Wetlabs ECO instrument).


ggalibert:
Note that ECO instruments can collect data in burst mode or not. One way to identify a bursting dataset is to look for the global attribute `instrument_burst_interval`.

mhidas:
Thanks @ggalibert 
Yes, I think we should use the presence of `instrument_burst_interval` and `instrument_burst_duration` to identify burst-sampled data sets.

PR #756, opened 2017-10-09
Pipeline v1 -> v2 migration

lbesnard:


lbesnard:
NOT TO BE MERGED

FOR RC pipeline 2 release. 
Some of my scripts need to be modified to work with the V2. This branch could be the one used for RC provisioning.
@mhidas @lwgordonimos @anguss00

mhidas:
Isn't this jumping the gun a bit? Wouldn't it be make more sense to deploy just a one or two new pipelines first and make sure it's all running ok before flicking the switch on a whole bunch of them? As far as I can remember that was the plan.

lbesnard:
We will be testing things on the new RC. Argo, FAIMMS , ANMN NRS AIMS, AUV, SOOP XBT DM need to have the code changed in the data services repo at the same time, even for testing, as they create what goes into the incoming dir

PR #762, opened 2017-10-16
We want to include any file with a data code T except those from ADCP…

ggalibert:
…s and CTD profiles.

### DIFF
--- a/ANMN/temperature_grid_product/anmn_temp_gridded_product.py
+++ b/ANMN/temperature_grid_product/anmn_temp_gridded_product.py
@@ -513,9 +513,7 @@ def main(incoming_file_path, deployment_code, output_dir, plot_comparaison=False
     s3_bucket_prefix = 'https://s3-ap-southeast-2.amazonaws.com/imos-data'
     logging           = IMOSLogging()
     logger            = logging.logging_start(os.path.join(os.environ['WIP_DIR'], 'anmn_temp_grid.log'))
-    list_fv01_url     = wfs_request_matching_file_pattern('anmn_all_map', '%%Temperature/%%_TZ_%%_FV01_%s%%' % deployment_code, s3_bucket_url=True)
-    list_fv01_url.extend(wfs_request_matching_file_pattern('anmn_all_map', '%%Temperature/%%_TE_%%_FV01_%s%%' % deployment_code, s3_bucket_url=True))
-    list_fv01_url.extend(wfs_request_matching_file_pattern('anmn_all_map', '%%Temperature/%%_TPE_%%_FV01_%s%%' % deployment_code, s3_bucket_url=True))
+    list_fv01_url     = wfs_request_matching_file_pattern('anmn_ts_timeseries_map', '%%_FV01_%s%%' % deployment_code, s3_bucket_url=True)
     previous_fv02_url = wfs_request_matching_file_pattern('anmn_all_map', '%%Temperature/gridded/%%_FV02_%s_%%gridded%%' % deployment_code)
 
     if len(previous_fv02_url) == 1:


bpasquer:
@ggalibert   just 2 comments- Originally the product was only using data from the anmn TS timeseries data sets with Is why I was only looking for files in 'Temperature/' folders. Secondly, the original regex excluded files that did not store a depth or pressure variable because I wanted to use only the actual depth as opposed to the nominal depth for the depth level of the instrument ( see for example http://thredds.aodn.org.au/thredds/catalog/IMOS/ANMN/QLD/GBRCCH/Temperature/catalog.html?dataset=IMOS/ANMN/QLD/GBRCCH/Temperature/IMOS_ANMN-QLD_T_20070910T073000Z_GBRCCH_FV01_GBRCCH-0709-SBE39-7_END-20080506T083500Z_C-20120206T043415Z.nc) 
You might want to look whether the processing script can successfully process such deployments and twick it to use the global attribute instrument_nominal_depth in these particular cases.

ggalibert:
> You might want to look whether the processing script can successfully process such deployments and twick it to use the global attribute instrument_nominal_depth in these particular cases.

Ok, will lodge a separate issue for that.

ANMN/temperature_grid_product/anmn_temp_gridded_product.py
```diff
@@ -513,9 +513,14 @@ def main(incoming_file_path, deployment_code, output_dir, plot_comparaison=False
     s3_bucket_prefix = 'https://s3-ap-southeast-2.amazonaws.com/imos-data'
     logging           = IMOSLogging()
     logger            = logging.logging_start(os.path.join(os.environ['WIP_DIR'], 'anmn_temp_grid.log'))
-    list_fv01_url     = wfs_request_matching_file_pattern('anmn_all_map', '%%Temperature/%%_TZ_%%_FV01_%s%%' % deployment_code, s3_bucket_url=True)
-    list_fv01_url.extend(wfs_request_matching_file_pattern('anmn_all_map', '%%Temperature/%%_TE_%%_FV01_%s%%' % deployment_code, s3_bucket_url=True))
-    list_fv01_url.extend(wfs_request_matching_file_pattern('anmn_all_map', '%%Temperature/%%_TPE_%%_FV01_%s%%' % deployment_code, s3_bucket_url=True))
+    list_fv01_url     = wfs_request_matching_file_pattern('anmn_all_map', '%%_%%T%%_%%_FV01_%s%%' % deployment_code, s3_bucket_url=True)
```
mhidas: I think that pattern needs to be a bit more specific. In particular `'%%_%%T%%_%%_'` will match all files because there's a 'T' in the timestamp.
Not sure what is possible with the patterns used by WFS. With regex you could do something like 
```regex
IMOS_ANMN-[A-Z]+_[^VW_]*T[^VW_]*_.*_FV01
```

mhidas:
> Originally the product was only using data from the anmn TS timeseries data sets

Actually @bpasquer that's the right approach, but that doesn't mean only the "Temperature" folders. In fact it doesn't matter what the url is, the selection has already been done by grouping those files into that collection. So instead of querying the 'anmn_all_map' WFS, the script should get the list of input files directly from 'anmn_ts_timeseries_map'!

bpasquer:
My comment was more about putting facts in their original context. There is now reason why we shouldn't expand the product to any kind of valid temperature measurements. Now obviously you've got a better knowledge of this collection to say whether valid data are available in the 'anmn_all_map' or in the 'anmn_ts_timeseries_map'

mhidas:
Yes, so what I'm saying is that all valid temperature measurements *should* be in the ANMN TS collection. This now *excludes* temperature measured by ADCP instruments, but *includes* all temperature loggers and CTD type instruments.

Oh, just remembered ANMN TS currently does not include temperature from WQM and other burst-sampling instruments. If we do include these, should we take the burst-averaged product (FV02) as input to the gridding, or unbinned (FV01) versions?

ggalibert:
> If we do include these, should we take the burst-averaged product (FV02) as input to the gridding, or unbinned (FV01) versions?

I suggest the unbinned (FV01) version since we're still having trouble with keeping binned (FV02) version up to date, yet the T data in the FV01 is very usable.

ggalibert:
> > If we do include these, should we take the burst-averaged product (FV02) as input to the gridding, or unbinned (FV01) versions?

> I suggest the unbinned (FV01) version since we're still having trouble with keeping binned (FV02) version up to date, yet the T data in the FV01 is very usable.

Actually, the reason the FV01 WQM are out of the database might be because there is a lot of data... Don't know if it's a good idea to try to push all that in the DB.

ggalibert:
@mhidas I have just updated the PR to take into account your comment and only list files from ANMN TS. At term it would be good to include FV02 burst averaged products T data in this collection too.

PR #765, opened 2017-10-20
ABOS_SOTS: skip buggy CF check_reduced_horizontal_grid

mhidas:
There's a bug in v2.x of the checker that causes it to incorrectly identify certain files as having reduced_horizontal_grid structure and reporting incorrect errors (see https://github.com/ioos/compliance-checker/issues/195 and https://github.com/ioos/compliance-checker/issues/284). They seem to have fixed it in v3, but we're not using that yet, so until then we're just going to skip this check.

So far this has only been a problem in this pipeline, so I'm fixing it here. If others have had this problem, we could apply the same fix centrally in https://github.com/aodn/data-services/blob/master/lib/common/netcdf-check.sh

### DIFF
--- a/ABOS/common/incoming_handler.sh
+++ b/ABOS/common/incoming_handler.sh
@@ -23,6 +23,38 @@ regex_filter() {
 }
 
 
+# trigger netcdf checker for file - SKIP CF check_reduced_horizontal_grid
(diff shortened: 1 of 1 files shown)


mhidas:
@smancini FYI

PR #769, opened 2017-10-31
Merge production commits back into master branch

anguss00:
@lwgordonimos @lbesnard fixes jenkins pipeline issue

### DIFF
--- a/SOOP/SOOP_BA/helper.py
+++ b/SOOP/SOOP_BA/helper.py
@@ -45,20 +45,24 @@ def open_dataset(nc_file, mode):
 
 def get_reporting_id(nc_file):
     F = open_dataset(nc_file, 'r')
-    code = get_code_long(nc_file)
-    date_start = getattr(F, 'time_coverage_start', '')
-    date_end = getattr(F, 'time_coverage_end', '')
-    F.close()
+    reporting_id = getattr(F, 'reporting_id', '')
+    if reporting_id:
+        return reporting_id
+    else:
+        code = get_code_long(nc_file)
+        date_start = getattr(F, 'time_coverage_start', '')
+        date_end = getattr(F, 'time_coverage_end', '')
+        F.close()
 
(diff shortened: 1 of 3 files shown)


mhidas:
What was this all about?? I thought those commits had been cherry-picked from master into production in the first place. They should have been in there already.

Also, it was not a good idea to delete the the **production** branch, because 10-aws-syd is now failing to provisison!

anguss00:
Yeah good question @mhidas. I am totally unsure how those commits got directly into the production branch, because they definitely weren't in the master branch. Perhaps they were cherry-picked directly from a feature branch? Cherry-picking must be handled carefully as there is no protection against pulling in feature branch commits. Regardless, we shouldn't be cherry-picking across data-services anyway, as we haven't got the issues we have in geoserver-config (and we don't want to be managing data services in the same way if at all possible)

lwgordonimos:
Actually, both of the commits *were* cherry-picked from master... These are the commits in question:

https://github.com/aodn/data-services/commit/859d1a6bb4c780ab64230377c1942aedfba1ddfc
https://github.com/aodn/data-services/commit/681e2c68cc4f2fb4d1f30c499c33ee85073def2f

They now appear to be duplicated from the reverse production->master PR...

https://github.com/aodn/data-services/commit/4bef54abff10488ac52c43cf6a7e48377880cee5
https://github.com/aodn/data-services/commit/5968743eef27bfad664dbc73f5b3e90015e65633

So:

a) the initial problem was my bad, because I thought data-services was the repo we currenrtly had to do the cherry picking on (which is actually geoserver-config)
b) we didn't actually need the PR back onto master
c) I have no idea why it wouldn't merge if those commits were there
d) we need to stop the cherry picking madness on geoserver-config to avoid confusion!

anguss00:
Ahh sorry mate my bad, was checking on sha basis. I wonder why on earth that prod merge failed if they were in both branches???

PR #778, opened 2017-11-23
Added check_netcdf_print_error to lib/common/netcdf-check.shlib/commo…

ggalibert:
…n/netcdf-check.sh so that the error message from ncdump can be retrieved and logged. Updated ACORN/BASH/incoming_handler_common.sh to take advantage of this new function.

### DIFF
--- a/ACORN/BASH/incoming_handler_common.sh
+++ b/ACORN/BASH/incoming_handler_common.sh
@@ -111,7 +111,8 @@ main() {
     [ -z "$ACORN_HOURLY_AVG_DIR" ] && file_error "missing ACORN_HOURLY_AVG_DIR variable"
 
     regex_filter $file || file_error "Did not pass ACORN regex filter"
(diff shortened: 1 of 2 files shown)


lwgordonimos:
Ok, had to revert this one.

Adding the "local" to the variables causes it to silently succeed in every case, because of some ridiculous bash quirk.

http://tldp.org/LDP/abs/html/localvar.html

> 

  | ... when declaring and               setting a local variable in a single command, apparently the               order of operations is to first set the variable,               and only afterwards restrict it to local scope.
-- | --


```bash
check_netcdf_print_error () 
{ 
    local file=$1;
    shift;
    local stderr;
    local exit_status;
    stderr=$(ncdump -h $file 2>&1 >/dev/null);
    exit_status=$?;
    [ $exit_status -ne 0 ] && echo "Failed netcdf format check with error: '$stderr'";
    return $exit_status
}
```
Declaring the variables as local first, before assigning them separately fixes it!

PR #788, opened 2017-12-01
Processing of NRL Adapter BY ANFOG DM pipeline

bpasquer:
and improvement of incoming with better use of local functions


mhidas:
@smancini Does this need to be merged & deployed before @bpasquer returns?

PR #792, opened 2017-12-19
Compliance Checker V3.1 - testing

lbesnard:
Not to be merged. Testing only


lib/python/cc_v3.1_testing.py
```diff
@@ -0,0 +1,69 @@
...
+"""
+
+import re
+import sys
+
+from util import pass_netcdf_checker
+from netCDF4 import Dataset
+
+
+def list_nc_checks_to_perform(netcdf_file_path):
```
mhidas: This should always return 'cf:1.6' and 'imos:1.4' unless the `Conventions` attribute exists and lists 'IMOS-1.3'.
In that case it should return 'cf:1.6' and 'imos:1.3'.

Issue #781, opened 2017-11-27
burst-averaging product tweaks 2


mhidas:
A few more issues with the code to create burst-averaged products:

* Some files legitimately don't have a `TEMP` variable (as mentioned [here](https://github.com/aodn/data-services/issues/746#issuecomment-332690626)). The [code](https://github.com/aodn/data-services/blob/master/ANMN/burst_averaged_product/burst_average.py#L235) assumes they always do, which means it breaks for these files. It shouldn't need to make this assumption.

* As far as I can tell, the code assumes that any gap between consecutive measurements of more than 1 second indicates a new burst (see [here](https://github.com/aodn/data-services/blob/master/ANMN/burst_averaged_product/burst_average.py#L162)) This may be true in most cases, but ideally this minimum gaps size should be set from the global attribute `instrument_sample_interval`.

mhidas:
Fixed in #794

PR #794, opened 2018-01-08
ANMN - burst average - fix issues #781

lbesnard:
* modify code to be more generic and not relying on TEMP to exist
 * using instrument_sample_interval global att value when exist

### DIFF
--- a/ANMN/burst_averaged_product/burst_average.py
+++ b/ANMN/burst_averaged_product/burst_average.py
@@ -217,7 +217,7 @@ def generate_netcdf_burst_filename(input_netcdf_file_path, burst_vars):
     pattern           = re.compile("^(IMOS_.*)_([0-9]{8}T[0-9]{6}Z)_(.*)_(FV0[0-9])_(.*)_END")
     match_group       = pattern.match(input_netcdf_name)
 
-    time_burst_vals = burst_vars[burst_vars.keys()[0]]['time_mean']
+    time_burst_vals = burst_vars.values()[0]['time_mean']
     time_min        = num2date(time_burst_vals, netcdf_file_obj['TIME'].units, netcdf_file_obj['TIME'].calendar).min().strftime('%Y%m%dT%H%M%SZ')
     time_max        = num2date(time_burst_vals, netcdf_file_obj['TIME'].units, netcdf_file_obj['TIME'].calendar).max().strftime('%Y%m%dT%H%M%SZ')
     burst_filename  = "%s_%s_%s_FV02_%s-burst-averaged_END-%s_C-%s.nc" % (match_group.group(1), time_min, \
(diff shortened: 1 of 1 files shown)


lbesnard:
related to https://github.com/aodn/data-services/issues/781

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -159,8 +159,10 @@ def burst_average_data(time_values, var_values, var_qc_exclusion):
     # look for start index of each burst, ie the index of each new spike of
     # difft variable
     idx_spike = [0] # initialise first index
-    spikes    = np.where(difft > 1)[0] + 1
+    spikes    = np.where(difft > INSTRUMENT_SAMPLE_INTERVAL)[0] + 1
     idx_spike.extend(spikes)
+    if idx_spike == [0]:
+        raise Exception('No burst was found in time values - FV01 is probably already averaged')
```
mhidas: A neater way to say this is:
```python
assert idx_spike == [0], 'No burst was found in time values'
```

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -215,7 +217,7 @@ def generate_netcdf_burst_filename(input_netcdf_file_path, burst_vars):
     pattern           = re.compile("^(IMOS_.*)_([0-9]{8}T[0-9]{6}Z)_(.*)_(FV0[0-9])_(.*)_END")
     match_group       = pattern.match(input_netcdf_name)
 
-    time_burst_vals = burst_vars['TEMP']['time_mean']
+    time_burst_vals = burst_vars[burst_vars.keys()[0]]['time_mean']
```
mhidas: You can just say
```python
time_burst_vals = burst_vars.values()[0]['time_mean']
```
(... but neither of these will work in Python 3)

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -231,8 +233,15 @@ def create_burst_average_netcdf(input_netcdf_file_path, output_dir):
     """
     input_file_rel_path = get_input_file_rel_path(input_netcdf_file_path)
     input_netcdf_obj    = Dataset(input_netcdf_file_path, 'r')
+
+    global INSTRUMENT_SAMPLE_INTERVAL
+    if hasattr(input_netcdf_obj, 'instrument_sample_interval'):
+        INSTRUMENT_SAMPLE_INTERVAL = input_netcdf_obj.instrument_sample_interval
+    else:
+        INSTRUMENT_SAMPLE_INTERVAL = 1
```
mhidas: You can replace the above 4 rows with:
```python
INSTRUMENT_SAMPLE_INTERVAL = getattr(input_netcdf_obj, 'instrument_sample_interval', 1)
```

ANMN/burst_averaged_product/burst_average.py
```diff
@@ -231,8 +233,15 @@ def create_burst_average_netcdf(input_netcdf_file_path, output_dir):
...
+
     burst_vars          = create_burst_average_var(input_netcdf_obj)
-    time_burst_vals     = burst_vars['TEMP']['time_mean']
+    time_burst_vals     = burst_vars[burst_vars.keys()[0]]['time_mean']
```
mhidas: as above...

ANMN/burst_averaged_product/burst_average.py
bpasquer: If you want to use an assertion a you should write 

    assert idx_spike != [0], 'No burst was found in time values'

since the assert function check that the statement is true and throw the error if not

lbesnard: good spot !

lwgordonimos: You shouldn't use assert for error handling in place of exceptions, they perform a different task.

https://wiki.python.org/moin/UsingAssertionsEffectively

```Assertions should *not* be used to test for failure cases that can occur because of bad user input or operating system/environment failures, such as a file not being found. Instead, you should raise an exception, or print an error message, or whatever is appropriate. One important reason why assertions should only be used for self-tests of the program is that assertions can be disabled at compile time. ```

If the user supplied input file is ultimately the source of the error, then it counts as "bad user input".

So if you're not happy for the assert statement to be stripped out of the code and have it still *fully functional*, it should be an explicit if statement and standard exception.

bpasquer: Thanks for clarifying that @lwgordonimos

mhidas: Ah, good point. Thanks.

lbesnard:
This is ready to go

PR #795, opened 2018-01-08
Temp Regridded - fix #699

### LINKED ISSUE #699: anmn_temp_gridded_product.py - Averaging process
I anticipate the current averaging process being an issue for users.

Ideally, one would like to retrieve averaged values that can fall on minute 00, regardless of when the deployment starts.

So, if a deployment starts at 10:06 and we want to perform 30min averages, then our first averaging cell should have its middle falling on 10:00, hence starts at 9:45 and finishes at 10:15. It is likely that for this first cell, there is not enough samples to meet the minimum required in the algorithm, but this doesn't matter.

lbesnard:
@ggalibert  https://github.com/aodn/data-services/issues/699

### DIFF
--- a/ANMN/temperature_grid_product/anmn_temp_gridded_product.py
+++ b/ANMN/temperature_grid_product/anmn_temp_gridded_product.py
@@ -172,11 +172,39 @@ def _perdelta(start, end, delta):
 
 def create_time_1d(time_start, time_end, delta_in_minutes):
     """
-    create a 1D time array between start and end date and a data step of delta_in_minute
+    create a 1D time array between start and end date and a data step of delta_in_minutes
     """
     time_interp_array = []
-    for result in _perdelta(time_start + timedelta(minutes=delta_in_minutes)/2,
-                            time_end - timedelta(minutes=delta_in_minutes)/2,
+
+    # we modify the time_start and time_end values
+    if delta_in_minutes == 30:
+        if time_start.minute <= 15:
+            time_start = time_start.replace(minute=0, second=0, microsecond=0)
+        elif time_start.minute > 15 and time_start.minute <= 45:
+            time_start = time_start.replace(minute=30, second=0, microsecond=0)
+        elif time_start.minute > 45:
+            time_start = time_start.replace(minute=0, second=0, microsecond=0) + timedelta(hours=1)
+
+        if time_end.minute <= 15:
+            time_end = time_end.replace(minute=0, second=0, microsecond=0)
+        elif time_end.minute > 15 and time_start.minute <= 45:
+            time_end = time_end.replace(minute=30, second=0, microsecond=0)
+        elif time_end.minute > 45:
+            time_end = time_end.replace(minute=0, second=0, microsecond=0) + timedelta(hours=1)
+
+    elif delta_in_minutes == 60:
+        if time_start.minute > 30:
+            time_start = time_start.replace(minute=0, second=0, microsecond=0) + timedelta(hours=1)
+        elif time_start.minte < 30:
+            time_start = time_start.replace(minute=0, second=0, microsecond=0)
+
+        if time_end.minute > 30:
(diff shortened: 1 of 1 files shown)


ANMN/temperature_grid_product/anmn_temp_gridded_product.py
```diff
@@ -512,8 +529,8 @@ def main(incoming_file_path, deployment_code, output_dir, plot_comparaison=False
     global fv01_dir
     s3_bucket_prefix = 'https://s3-ap-southeast-2.amazonaws.com/imos-data'
     logging           = IMOSLogging()
-    logger            = logging.logging_start(os.path.join(os.environ['WIP_DIR'], 'anmn_temp_grid.log'))
-    list_fv01_url     = wfs_request_matching_file_pattern('anmn_ts_timeseries_map', '%%_FV01_%s%%' % deployment_code, s3_bucket_url=True)
+    logger            = logging.logging_start(os.path.join(output_dir, 'anmn_temp_grid.log'))
+    list_fv01_url     = wfs_request_matching_file_pattern('anmn_ts_timeseries_map', '%%_FV01_%s%%' % deployment_code, s3_bucket_url=True, url_column='file_url')
```
ggalibert: These two changes above should go in a separate commit / PR.

ggalibert:
We need a different logic since this is only going to work for a dataset that needs averaging over 30min, however the averaging time period is dependant on the site sampling regime (see [`get_frequency_step_in_deployment`](https://github.com/aodn/data-services/blob/98ebf9631889903561af3b47f0ec046e6f2e6f92/ANMN/temperature_grid_product/anmn_temp_gridded_product.py#L198)).

lbesnard:
ok will think about this tomorrow

lbesnard:
@ggalibert done some mods. I removed the 90' option. I think  using 120'  average could be a bit annoying.

ANMN/temperature_grid_product/anmn_temp_gridded_product.py
```diff
@@ -172,11 +172,39 @@ def _perdelta(start, end, delta):
...
+    elif delta_in_minutes == 60:
+        if time_start.minute > 30:
+            time_start = time_start.replace(minute=0, second=0, microsecond=0) + timedelta(hours=1)
+        elif time_start.minte < 30:
+            time_start = time_start.replace(minute=0, second=0, microsecond=0)
+
+        if time_end.minute > 30:
+            time_end = time_end.replace(minute=0, second=0, microsecond=0) + timedelta(hours=1)
+        elif time_end.minute < 30:
+            time_end = time_end.replace(minute=0, second=0, microsecond=0)
```
ggalibert: There is a more generic way you can do this for any delta_in_minutes value. Your new time_start for example could be set by something like:
```Matlab
time_start_in_minutes = floor(time_start_in_minutes/delta_in_minutes)*delta_in_minutes
```

ANMN/temperature_grid_product/anmn_temp_gridded_product.py
```diff
@@ -229,8 +257,8 @@ def _sample_inter(sample_interval):
             average_window = 60
             n_val_step     = 3
         elif round(sample_interval) >= 1200:
-            average_window = 90
-            n_val_step     = 3
+            average_window = 60
+            n_val_step     = 2
```
ggalibert: Why 2?

lbesnard: it's the amount of valid measures (good qc) we need in the average window to perform a average. see with @bpasquer

ggalibert:
Closing this since the following PR https://github.com/aodn/data-services/pull/809 includes some of the above modifications + a fix to https://github.com/aodn/data-services/issues/699 with the following commit: https://github.com/aodn/data-services/pull/809/commits/0e0cf84fb3a469d2ef540fdd2edfcd82bc8052e4.

PR #796, opened 2018-01-15
Argo - force rsync log files to only contain uniq lines

lbesnard:


ARGO/rsync.sh
```diff
@@ -24,7 +24,7 @@ main() {
     fi
 
     local tmp_rsync_output_file=`mktemp`
-    rsync_argo | tee $tmp_rsync_output_file
+    rsync_argo | sort | uniq | tee $tmp_rsync_output_file
```
mhidas: FYI `sort -u` does the same thing.

PR #805, opened 2018-02-08
AUV - migration to pipeline v2 - renaming manifest files

lbesnard:


lbesnard:
linked to https://github.com/aodn/chef-private/pull/2687

PR #807, opened 2018-02-13
Faimms anmn dar pipelin2

lbesnard:


lbesnard:
@mhidas thanks for pointing this out. 
The explanation is because the main faimms script uses some routines in dest_path to grab some information, such as the main variable of a file ...
see https://github.com/aodn/data-services/blob/363e7494a8924f3d736f29f281f6b91c9fff6b8f/FAIMMS/REALTIME/faimms.py#L37

This is why there was a commit to re add the dest_path for faimms. However this makes me realise now that the same need to happen for the ANMN one.

mhidas:
@lbesnard No worries. Might be nice to sqash some of these commits to keep the history a bit cleaner.

lbesnard:
@mhidas fixed and now pretty ! ready to be merged and tested on Nec 4 
https://github.com/aodn/chef-private/pull/2693

PR #810, opened 2018-02-14
SOOP XBT DM - handle case when data is not masked

lbesnard:


lbesnard:
To be tested by Rebecca

lbesnard:
tested  and all good

PR #814, opened 2018-02-28
AUV - fix umask to 0o002 to be 664 on files and 775 on folders. Previ…

lbesnard:
…ously was 0o112


lbesnard:
@lwgordonimos see http://wintelguy.com/umask-calc.pl

PR #816, opened 2018-03-02
Remove SOOP-CO2 watch before deployment of pipeline v2

bpasquer:


anguss00:
More than happy to merge, just have a few Qs about the chef PR

PR #818, opened 2018-03-06
put soop_co2 watch.d back

bpasquer:


bpasquer:
we need it back in prod until the pipeline v2 processing of realtime files is ready

PR #827, opened 2018-03-15
Revert "SRS GHRSST pip2 - remove unittest pipeline1"

lbesnard:
Reverts aodn/data-services#825

back to pipeline 1


lbesnard:
https://github.com/aodn/chef-private/pull/2739
https://github.com/aodn/data-services/pull/826

PR #831, opened 2018-04-05
T gridded improvements

ggalibert:


ANMN/temperature_grid_product/anmn_temp_gridded_product.py
```diff
@@ -430,7 +430,7 @@ def main(incoming_file_path, deployment_code, output_dir):
     
     try:
         vargs = args()
-        fv02_nc_path, previous_fv02_url = main(vargs.incoming_file_path, vargs.deployment_code, vargs.output_dir)
+        fv02_nc_path, previous_fv02_url, fv01_dir = main(vargs.incoming_file_path, vargs.deployment_code, vargs.output_dir)
         if fv02_nc_path is not None:
             print fv02_nc_path, previous_fv02_url
```
lbesnard: what is the purpose of this line ? If it is to output things to the terminal for logging purposes, you could use logger.info

ANMN/temperature_grid_product/anmn_temp_gridded_product.py
```diff
@@ -63,8 +63,8 @@ def get_good_values(var, var_qc):
     """
     Return the variable values which qc flag is 0, 1 or 2
     """
-    var = var[var_qc <= 2]
-        
+    var = [i for i,j in zip(var, var_qc) if j <= 2]
+
```
lbesnard: i actually don't care at all, but this extra empty line is not needed (see PEP8)

ANMN/temperature_grid_product/anmn_temp_gridded_product.py
ggalibert: It is so that I can retrieve these when calling the script from another one. For example in a BASH script:
```bash
output=`python /mnt/ebs/data-services/ANMN/temperature_grid_product/anmn_temp_gridded_product.py -d $deployment`
if [ -z "$output" ]; then
	echo "Not enough files to create product for $deployment" >> ~/processTGridded.log
else
	new_file=`echo $output | cut -d ' ' -f 1`
	old_file=`echo $output | cut -d ' ' -f 2`
fi
```

ANMN/temperature_grid_product/anmn_temp_gridded_product.py
ggalibert: Thank @lbesnard , if happy then could you please merge?

ANMN/temperature_grid_product/anmn_temp_gridded_product.py
lbesnard: maybe you should call python from python, rather than from bash. Would def help to prevent errors.

PR #846, opened 2018-05-22
SRS OC BODBAW - change generate_netcdf_att library to handle case whe…

lbesnard:
…re units = "1"

### DIFF
--- a/lib/python/generate_netcdf_att.py
+++ b/lib/python/generate_netcdf_att.py
@@ -131,6 +131,10 @@ def _setup_var_att(nc_varname, netcdf4_obj, parser, conf_file_point_of_truth):
     for var_attname, var_attval in var_atts.iteritems():
         setattr(var_object, var_attname, _real_type_value(var_attval))
 
+        # handle case when units is equal to 1 and needs to be a str
+        if var_attname == 'units' and var_attval == '1':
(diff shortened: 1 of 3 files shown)


lib/python/generate_netcdf_att.py
```diff
@@ -131,6 +131,10 @@ def _setup_var_att(nc_varname, netcdf4_obj, parser, conf_file_point_of_truth):
     for var_attname, var_attval in var_atts.iteritems():
         setattr(var_object, var_attname, _real_type_value(var_attval))
 
+        # handle case when units is equal to 1 and needs to be a str
+        if var_attname == 'units' and var_attval == '1':
```
mhidas: Could this be done in a more generic way, so that you can set any attribute in your config file to be string type?
What happens if the config file says `units = '1'`?

lbesnard: the config file should not be written this way

lbesnard: will have to revisit all of this in the future when we will adopt a common way of generating NCs

mhidas: Ok. So do you just want this merged as is for now?

lbesnard: I think for now it makes sense

PR #850, opened 2018-05-31
ACORN - remove pipeline 1 - unused code

lbesnard:


lbesnard:
linked to https://github.com/aodn/chef-private/pull/2842

ACORN/BASH/ncDiff.sh
```diff
@@ -1,50 +0,0 @@
...
-secondFilePath=${secondFile%/*}
-
-
-# dump the files
-#ncdump $option $firstFile > /tmp/$firstFileName.txt
-#ncdump $option $secondFile > /tmp/$secondFileName.txt
-#diff /tmp/$firstFileName.txt /tmp/$secondFileName.txt
-
-# let's do it in memory rather than writing to disk
-diff <(ncdump $option $firstFile) <(ncdump $option $secondFile)
```
ggalibert: Can we please leave this script there? I'm still using it every now and then.

lbesnard: weird that it was in this folder. but anyway it exists here
https://github.com/aodn/data-services/blob/master/lib/BASH/ncDiff.sh

PR #862, opened 2018-08-14
Revert "ACORN - remove pipeline 1 - unused code"

lbesnard:
Reverts aodn/data-services#850


ggalibert:
Due to new pipeline v2 not including error dir clearing feature after successful re-upload (aodn/python-aodndata#74) and not throwing error/notification for upload of older file (aodn/python-aodndata#69).

PR #864, opened 2018-08-28
ANMN DAR YON - change INCOMING DIR PATH

lbesnard:


ANMN/NRS_AIMS/REALTIME/anmn_nrs_aims.py
```diff
@@ -49,6 +49,9 @@
 from util import pass_netcdf_checker
 
 
+ANMN_NRS_INCOMING_DIR = os.path.join(os.environ.get('INCOMING_DIR'), 'ANMN_NRS_DAR_YON')
```
mhidas: Please change `'ANMN_NRS_DAR_YON'` to `'ANMN', 'NRS_DAR_YON' ` if you don't mind. Thanks.

lbesnard: I don't mind. but  thought removing it from your ANMN directory would remove any confusion to your uploaders since they won't see it

mhidas: Good point! But maybe let's keep the root incoming directory clean too. We already have an [AODN](https://github.com/aodn/chef-private/blob/master/data_bags/ftp_dir_tree/AODN.json) directory there, so maybe make it "AODN/NRS_DAR_YON" ("AODN/moorings" is for data products published manually by us)

mhidas: Yeah, that's confusing. In this case "AODN" means us, the IMOS facility, because it's for our internal use.
(We already have separate base directories for non-IMOS "AODN" data e.g. CSIRO and NSW-OEH)

PR #867, opened 2018-08-30
update qc_flags file

bpasquer:


bpasquer:
Would you mind merging and deploying to RC please @lbesnard

PR #882, opened 2018-11-07
AODN WAVE DM: update standard name vocabulary global attribute to the…

bpasquer:
… latest CF standartd name table number


bpasquer:
Should have been updated to the latest version, and should help with this issue see the comment https://github.com/aodn/content/issues/362#issuecomment-436498282

PR #889, opened 2018-11-30
AIMS script - major refactoring - add retry on downloads - bug fix

lbesnard:


lbesnard:
old pickles files not compatible

PR #905, opened 2019-02-20
Update chlorophyll parameters based on fluorometers

mhidas:
This adds variables previously missing from the CSV metadata headers, and specifies a more accurate long_name for chlorophyll variables based on florometers.

There is currently no standard_name or imos_vocabulary_name for these!


mhidas:
I noticed that SOOP TMV and AUV collections also refer to `CPHL`. Are these variables also derived from florometers? In that case they should also be updated. @bpasquer @lbesnard

mhidas:
@bpasquer How do you test these changes? Only in RC? The parameters_mapping harvester doesn't seem to be set up in the PO box.

bpasquer:
Testing is done on RC. I tried to set up the parameters_mapping harvester in my box in the past but it appeared to be too much hassle.

mhidas:
:+1: Thanks

mhidas:
These will need to be updated again with a new IMOS parameter name when https://github.com/aodn/content/issues/403 is done.

bpasquer:
From a user perpective, I think you need to define three different parameters here. You can advantage of the fact that you define these variables using the long name which gives more latitude in the way we define parameters to provide more details about what distinguishes them for each other.

ggalibert:
> I noticed that SOOP TMV and AUV collections also refer to `CPHL`. Are these variables also derived from florometers? In that case they should also be updated.

Yes, @bpasquer @lbesnard please open a content issue on this for each the collections mentioned above.

mhidas:
> From a user perpective, I think you need to define three different parameters here.

I agree that it would be helpful for the user to have some information about the difference between the three variables. However, I don't think there's currently enough flexibility here to provide this. The three variables are actually the same physical property, derived the same way, so they have the same long_name and the same units. The only difference is the source of the coefficents used in converting fluorescence to chlorophyll (and it seems we're not even sure there's really a difference!) So if we were going to provide this extra detail, it would just be a "comment".

@ggalibert What do you think?

ggalibert:
> So if we were going to provide this extra detail, it would just be a "comment".

Agreed, this could be an improvement to be made in the toolbox. See https://github.com/aodn/imos-toolbox/issues/532 .

bpasquer:
But there is no "comment" field in the CSV metadata header.

ggalibert:
> But there is no "comment" field in the CSV metadata header.

Agreed, CSV sucks a bit ;-) . Because of that and for like many other metadata (conversion coefficients??!), I don't think it's worth trying to include this info in the CSV.

bpasquer:
Then I suggest we introduce a field "comment " in the parameters.csv file and modify the harvester.

mhidas:
Never mind the comments, I just realised the "long_name" in parameters.csv is not actually included in the metadata header, so at the moment the only information we get for the chlorophyll parameters is the units. Luckily it seems there _is_ already an IMOS vocab name for it (https://github.com/aodn/content/issues/403#issuecomment-465860910 - thanks @bpasquer!)

bpasquer:
In order to add a particular field (eg long_name in this case) in the metadata header,  you need to modify the data collection specific views in the paramers_mapping harvester. See https://github.com/aodn/harvesters/pull/679

PR #908, opened 2019-03-04
add full description of chlorophyll parameters using comment

bpasquer:


bpasquer:
Associated to https://github.com/aodn/harvesters/pull/685 to fix aodn/content#404

mhidas:
:+1: Thanks @bpasquer, looks good. We should add the IMOS vocab name as well. Currently it's "Concentration of inferred chlorophyll from relative fluorescence per unit volume of the water body". I _did_ suggest renaming it (https://github.com/aodn/content/issues/403), but that's a more long-term possibility, so we'll update it here if/when that happens.

Where does the `imos_vocabulary_id` come from? @atkinsn @smancini

PARAMETERS_MAPPING/parameters.csv
```diff
@@ -1,170 +1,172 @@
...
-167;;;sea_surface_swell_wave_significant_height;;
-168;;;sea_surface_swell_wave_period_at_variance_spectral_density_maximum;;
-169;sea_surface_swell_mean_period_from_variance_spectral_density_first_frequency_moment;;;;
-170;;;sea_surface_swell_wave_from_direction_at_variance_spectral_density_maximum;;
-171;;;sea_surface_wind_wave_significant_height;;
-172;;;sea_surface_wind_wave_period_at_variance_spectral_density_maximum;;
-173;sea_surface_wind_wave_mean_period_from_variance_spectral_density_first_frequency_moment;;;;
-174;;;sea_surface_wind_wave_from_direction_at_variance_spectral_density_maximum;;
-175;;;mass_concentration_of_inferred_chlorophyll_from_relative_fluorescence_units_in_sea_water;;
+unique_id;cf_standard_name;cf_alias_name;long_name;imos_vocabulary_name;imos_vocabulary_id,comment
```
mhidas: Separator is semi-colon, not comma, so should be `;comment` at the end.

mhidas:
Regarding the `imos_vocabulary_id` column, after having a bit of closer look (thanks @atkinsn), it seems these can no longer be used to automatically generate a URI for the vocab term (because several different id sources  have been combined). We should really replace  this column with an actual URI, like
http://vocab.nerc.ac.uk/collection/P01/current/TEMPPR01 (for Temperature of the water body), or 
http://vocab.aodn.org.au/def/discovery_parameter/894 (for Concentration of inferred chlorophyll from relative fluorescence per unit volume of the water body). Then we could even include this info in the CSV metadata header.

bpasquer:
Happy to do it, although a bit of help from @atkinsn would be appreciated ;-)
@ggalibert ?

bpasquer:
I fixed the issue you found @mhidas , so maybe  you can merge this PR  and open a separate issue about the addition of the URI in the metadata header

atkinsn:
Can do @bpasquer

PARAMETERS_MAPPING/parameters.csv
```diff
@@ -1,170 +1,172 @@
...
+168;;;sea_surface_swell_wave_period_at_variance_spectral_density_maximum;;;
+169;sea_surface_swell_mean_period_from_variance_spectral_density_first_frequency_moment;;;;;
+170;;;sea_surface_swell_wave_from_direction_at_variance_spectral_density_maximum;;;
+171;;;sea_surface_wind_wave_significant_height;;;
+172;;;sea_surface_wind_wave_period_at_variance_spectral_density_maximum;;;
+173;sea_surface_wind_wave_mean_period_from_variance_spectral_density_first_frequency_moment;;;;;
+174;;;sea_surface_wind_wave_from_direction_at_variance_spectral_density_maximum;;;
+175;;;mass_concentration_of_inferred_chlorophyll_from_relative_fluorescence_units_in_sea_water;;conversion from fluorescence to chlorophyll-a using unknown coefficients
+176;;;mass_concentration_of_inferred_chlorophyll_from_relative_fluorescence_units_in_sea_water;;conversion from fluorescence to chlorophyll-a using factory coefficients
+177;;;mass_concentration_of_inferred_chlorophyll_from_relative_fluorescence_units_in_sea_water;;conversion from fluorescence to chlorophyll-a using user coefficients
```
mhidas: The comments you added in the last three rows are actually in the `imos_vocabulary_id` column. You need one more semi-colon in there.

mhidas:
@bpasquer One more issue to fix. After that I would be happy to merge, though I think we may as well add in the imos_vocabulary_name in this PR, since it does exist.

atkinsn:
Can do @bpasquer

bpasquer:
I've  added the imos vocab param and fixed the other issue @mhidas

mhidas:
Cool, thanks!

PR #928, opened 2019-05-09
TSaggregator: fix issue running on Windows by closing Datasets while …

lwgordonimos:
…iterating over them

Out of morbid interest, I replicated the issue on my Windows VM and narrowed it down to the fact that the nc1 variable was being overwritten in the loop without closing the Dataset from the previous iteration. For some reason, this is a problem on Windows but not on Linux.

### DIFF
--- a/ANMN/LTSP/TSaggregator/copyDataset.py
+++ b/ANMN/LTSP/TSaggregator/copyDataset.py
@@ -388,23 +388,19 @@ for v in var_names_out:
         for path_file in files:
             #print("%d : %s file %s" % (filen, v, path_file))
             print("%s " % (filen+1), end="")
-            nc1 = Dataset(path_file, mode="r")
+            with Dataset(path_file, mode="r") as nc1:
 
+                n_records = len(nc1.dimensions['TIME'])
 
-            n_records = len(nc1.dimensions['TIME'])
-
-            ## EK. check if the variable is present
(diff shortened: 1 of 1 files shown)


mhidas:
Strange, there are actually other cases where a Dataset isn't closed:
https://github.com/aodn/data-services/blob/dea5e3975cfbc3c7482e8236071516cea3ff53c8/ANMN/LTSP/TSaggregator/copyDataset.py#L207 (not in loop)

https://github.com/aodn/data-services/blob/dea5e3975cfbc3c7482e8236071516cea3ff53c8/ANMN/LTSP/TSaggregator/copyDataset.py#L347 (in a loop!)

I don't know why the second case isn't causing the same problem, but wouldn't hurt to put that in a `with` clause as well?

lwgordonimos:
@mhidas it definitely wouldn't hurt, but I withstood the temptation to add it elsewhere and just targeted the specific problematic one. I always tend use context managers everywhere, because they tend to help prevent weird bugs like this further down the code.

diodon:
Thanks for looking at this @lwgordonimos. I closed the dataset inside the loop and now the colleague who reported the problem was able to run it smoothly in her windows machine.

diodon:
thanks @mhidas. I've fixed that as well.

mhidas:
> thanks @mhidas. I've fixed that as well.

Great! Can you please push that change on top of this branch? Than we can merge it in.

PR #929, opened 2019-05-10
T saggregator

diodon:
Better global attributes management. This version was used to generate the sample files ready for revision

### DIFF
--- a/ANMN/LTSP/TSaggregator/copyDataset.py
+++ b/ANMN/LTSP/TSaggregator/copyDataset.py
@@ -207,6 +207,7 @@ nc_time = ds_time.get_variables_by_attributes(standard_name='time')
 
 dates = num2date(ma_time_all[idx].compressed(), units=nc_time[0].units, calendar=nc_time[0].calendar)
 
+
 #
 # createNewFile
 #
@@ -320,6 +321,7 @@ if len(files) > 128:
     if len(files) > 32767: # your really keen then
         index_var_type = "i4"
 
+
 #
 # create new variables needed
 #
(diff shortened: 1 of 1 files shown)


ANMN/LTSP/TSaggregator/copyDataset.py
```diff
@@ -462,27 +477,41 @@
...
             if hasattr(nc_variable_out, 'standard_name'):
                 if nc_variable_out.standard_name == 'latitude':
                     la_max = ma_variable_all.max(0)
                     la_min = ma_variable_all.max(0)
-                    nc_out.setncattr("geospatial_lat_max", la_max)
-                    nc_out.setncattr("geospatial_lat_min", la_min)
+                    gattr_tmp.update({"geospatial_lat_max": la_max})
+                    gattr_tmp.update({"geospatial_lat_min": la_min})
+                    #nc_out.setncattr("geospatial_lat_max", la_max)
+                    #nc_out.setncattr("geospatial_lat_min", la_min)
```
mhidas: These comments can be  removed.

ANMN/LTSP/TSaggregator/copyDataset.py
```diff
@@ -462,27 +477,41 @@
...
+                    gattr_tmp.update({"geospatial_vertical_min": d_min})
+                    #nc_out.setncattr("geospatial_vertical_max", d_max)
+                    #nc_out.setncattr("geospatial_vertical_min", d_min)
 
 
-nc.close()
+# sort new global attr dictionary
+gattr_new={}
+for key, value in sorted(gattr_tmp.items()):
+    gattr_new.update({key: value})
```
mhidas: This doesn't achieve the desired effect in Python v<3.6 because dict items are not ordered. But you can do away with the intermediate dict and add them to the file directly:
```python
for key, value in sorted(gattr_tmp.items()):
    nc_out.setncattr(key, value)
```

mhidas:
I'm going to merge this for now. Further fixes and enhancements will be done in separate PRs.

mhidas:
For an example of how the aggregation fails when one of the input files has a missing variable, try e.g.
```
copyDataset.py -var TEMP -site NRSROT -ts 2014-12-01 -te 2015-01-01
```

PR #930, opened 2019-05-23
MHL SST: update watch to point to handler file new location

bpasquer:


bpasquer:
You should be able to trigger the MHL SST  handler with this PR @craigrose

PR #934, opened 2019-06-05
Geoserver catalog

diodon:
new function to harvest urls from geoserver with some arguments.


ANMN/LTSP/TSaggregator/test/geoserverCatalog.py
```diff
@@ -0,0 +1,98 @@
...
+import pandas as pd
+
+
+
+def args():
+    parser = argparse.ArgumentParser(description="Get a list of urls from the AODN geoserver")
+    parser.add_argument('-var', dest='varname', help='name of the variable of interest, like TEMP', required=True)
+    parser.add_argument('-site', dest='site', help='site code, like NRMMAI',  required=True)
+    parser.add_argument('-ft', dest='featuretype', help='feature type, default timeseries', default='timeseries', required=False)
+    parser.add_argument('-fv', dest='fileversion', help='file version 1 or 2 only', default=1, type=int, required=False)
```
mhidas: Same here, default should include all file versions. (and why does the help say  "version 1 or 2 only"?)

ANMN/LTSP/TSaggregator/test/geoserverCatalog.py
```diff
@@ -0,0 +1,98 @@
...
+    parser.add_argument('-ts', dest='timestart', help='start time like 2015-12-01. Default 1944-10-15', default='1944-10-15', required=False)
+    parser.add_argument('-te', dest='timeend', help='end time like 2018-06-30. Default today\'s date', default=str(datetime.now())[:10], required=False)
```
mhidas: And again. Ok, in this case what you have is pretty much equivalent in practice, but default should be no filtering on start or end time if not specified.

ANMN/LTSP/TSaggregator/test/geoserverCatalog.py
```diff
@@ -0,0 +1,98 @@
...
+    parser.add_argument('-out', dest='outFileList', help='name of the file to store the selected files urls. Default: filelist.csv', default="", required=False)
```
mhidas: The specified `default=""` seems inconsistent with the help text.

ANMN/LTSP/TSaggregator/test/geoserverCatalog.py
```diff
@@ -0,0 +1,98 @@
...
+    parser.add_argument('-realtime', dest='realtime', help='real time data wanted?', required=False)
+
+    vargs = parser.parse_args()
+    return(vargs)
+
+
+def get_urls(varname, site, featuretype='timeseries', fileversion=1, realtime='False', timestart='1944-10-15', timeend=str(datetime.now())[:10], outFileList=''):
```
mhidas: All the comments I made about default values apply here too. I would make every arg `=None` by default. Then in the code you can can apply each filter only if it was given.
e.g.
```python
if site is None:
    filter_xml = None
else:
    filter_site = PropertyIsLike(propertyname='site_code', literal=site)
    filter_xml = etree.tostring(filter_site.toXML()).decode("utf-8")

response = wfs.getfeature(typename=ftype, filter=filter_xml,  outputFormat='csv')
```

ANMN/LTSP/TSaggregator/test/geoserverCatalog.py
```diff
@@ -0,0 +1,98 @@
...
+
+    ## filter first by site
+    filter_site = PropertyIsLike(propertyname='site_code', literal=site)
+    filter_xml = etree.tostring(filter_site.toXML()).decode("utf-8")
+    response = wfs.getfeature(typename=ftype, filter=filter_xml,  outputFormat='csv')
+
+    df = pd.read_csv(response)
+
+    ## then filter by the rest of criteria
+    criteria_var = df.variables.str.contains(varname, regex=False)
```
mhidas: Have to be careful with sub-string matching here. E.g. if varname is "TEMP", you should be looking for sub-string " TEMP," otherwise you might also get variables with names like "XTEMP" or "TEMP_SOMETHING_ELSE".

ANMN/LTSP/TSaggregator/test/geoserverCatalog.py
```diff
@@ -0,0 +1,98 @@
...
+    criteria_fv = df.file_version == fileversion
+    try:
+        date_start = datetime.strptime(timestart, '%Y-%m-%d')
+        date_end = datetime.strptime(timeend, '%Y-%m-%d')
+    except ValueError:
+        sys.exit('ERROR: invalid start or end date.')
+
+    criteria_startdate = pd.to_datetime(df.time_coverage_start) >= date_start
+    criteria_enddate = pd.to_datetime(df.time_coverage_end) <= date_end
```
mhidas: These are the wrong way around again. If I specify a start date, I want all files that have _any data_ after that date, i.e. those with _time_coverage_end_ after the specified date. And vice-versa for end date.

ANMN/LTSP/TSaggregator/test/geoserverCatalog.py
```diff
@@ -0,0 +1,98 @@
...
+
+    if realtime.lower()=='true':
```
mhidas: Why not make `realtime` a boolean? (or `None` for no filtering)

ANMN/LTSP/TSaggregator/test/geoserverCatalog.py
```diff
@@ -0,0 +1,98 @@
...
+        criteria_realtime = df.realtime==True
+        criteria_all = criteria_var & criteria_realtime & criteria_fv & criteria_startdate & criteria_enddate
+    else:
+        criteria_realtime = df.realtime==False
+        criteria_ft = df.feature_type.str.lower() == featuretype
+        criteria_all = criteria_var & criteria_ft & criteria_realtime & criteria_fv & criteria_startdate & criteria_enddate
```
mhidas: The logic connecting `realtime` and `featuretype` filters doesn't belong here. Just apply them independently.

ANMN/LTSP/TSaggregator/test/geoserverCatalog.py
```diff
@@ -0,0 +1,98 @@
...
+
+    if outFileList:
+        df.url[criteria_all].to_csv(outFileList, index=False)
+
+    return(df.url[criteria_all])
```
mhidas: I think this returns a `pandas.Series`. Let's convert it into a simple list before returning.

ANMN/LTSP/TSaggregator/test/geoserverCatalog.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+if __name__ == "__main__":
+    vargs           = args()
+    filenames = get_urls(**vars(vargs))
```
mhidas: Just need to print the list of urls, one per line!

ggalibert:
A quick comment on something that just got on my mind. This is not necessarily critical to deal with this right now but for the future be aware that it may be possible that some instruments have output a parameter twice (2 identical sensors or 2 ways of measuring the same thing) and the way it is recognised in the NetCDF file produced by the toolbox is when you find a dataset with a variable name say PARAM and then another PARAM_2 .

mhidas:
@ggalibert Yes, that is a possibility we'll need to handle at some stage.

ANMN/LTSP/TSaggregator/test/geoserverCatalog.py
mhidas: Another thought on this function, since it is hard-coded to query the "moorings_all_map" layer, and it wouldn't make sense to generalise it further,  the  function could be called `get_moorings_urls`.

ANMN/LTSP/TSaggregator/test/geoserverCatalog.py
```diff
@@ -0,0 +1,98 @@
...
+from owslib.fes import *
+from owslib.etree import etree
+import pandas as pd
+
+
+
+def args():
+    parser = argparse.ArgumentParser(description="Get a list of urls from the AODN geoserver")
+    parser.add_argument('-var', dest='varname', help='name of the variable of interest, like TEMP', required=True)
+    parser.add_argument('-site', dest='site', help='site code, like NRMMAI',  required=True)
```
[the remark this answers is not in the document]
diodon: working on it...

ANMN/LTSP/TSaggregator/test/geoserverCatalog.py
```diff
@@ -0,0 +1,98 @@
+## read all ANMN/ABOS files names form the geoserver
+"""
+geoserverCatalog.py
+
+Collect files names from the AODN geoserver according to several conditions
+Output a list of urls and optionally write into a text file
+
+usage: geoserverCatalog.py [-h] -var VAR -site SITE [-ft FEATURETYPE] [-fv FV]
```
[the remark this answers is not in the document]
diodon: removed

ANMN/LTSP/TSaggregator/test/geoserverCatalog.py
diodon: Yes this is a problem indeed. Not only for the extra chars after or before the variable name but with the position of the name in the `variables` string (', TEMP,' the most frequent, 'TEMP, 'if at the beginning, ', TEMP' if at the end. So comma as part of the search string does not solve the problem.

I found a solution using a regex with a word boundary in the Series.str.contains method:

`str.contains('.*\\bTEMP\\b.*', regex=True) `

ANMN/LTSP/TSaggregator/test/geoserverCatalog.py
diodon: The proposed way will use only full deployments in the date interval. But yes, you can have data in a deployment that partially match the date interval. Fixed

ocehugo:
Comments:

0. Add a pytest test. Example:

````
default_opts = { 'var': 'UCUR' , ... }

def test_geoserver_catalog():
     res = geoserver_catalog(default_opts)
     assert len(res) > 0
```

Now you can do continue or add more argument options to test other stuff ! Everytime you code something, just run pytest.


1. It would be nice to have a constructor for the geoserver html address. This will futureproof and be more generic in the construction of the query.

2. As said, have a look at docopt. It will take the argparse boilerplate out and help your "help".

3.  Collect the errors instead of raising them. We will need to iterate on the script several times if there is more than one error. An useful way to collect errors and raise exceptions is to use generators. 

5. Your functions are still too big. For example, if a generator function is used to check and raise Exceptions regarding fileversion,variable, timestart, etc - your function is greatly reduced and all checks are contained in one function.

6. I would try to be more idiomatic. Example:

```if timestart_is_valid
           raise Exception
```

7. Use file context: ``` with open( ... ```

With the above modifications your main function will have less than 10 lines, be very idiomatic, and testable.

ANMN/LTSP/TSaggregator/geoserverCatalog_generic.py
```diff
@@ -0,0 +1,112 @@
+## read all ANMN/ABOS files names form the geoserver
+#!/usr/bin/env python3
```
mhidas: The `#!` needs to be the _very first_ line of the file, followed by the coding. Doc strings and other comments can come after that.

ANMN/LTSP/TSaggregator/geoserverCatalog_generic.py
```diff
@@ -0,0 +1,112 @@
...
+Collect files names from the AODN geoserver according to several conditions
+Output a list of urls and optionally write into a text file
+
+"""
+
+
+import sys
+import argparse
+from datetime import datetime, timedelta
+from distutils.util import strtobool
```
mhidas: `timedelta` and `strtobool` are not used in the script, so don't need to be imported

diodon: remnant of previous version. Removed.

ANMN/LTSP/TSaggregator/geoserverCatalog_generic.py
diodon: fixed

ANMN/LTSP/TSaggregator/geoserverCatalog_generic.py
```diff
@@ -0,0 +1,112 @@
...
+
+    if realtime:
+        if realtime.lower() == "yes":
+            url = "http://geoserver-123.aodn.org.au/geoserver/ows?typeName=moorings_all_map&SERVICE=WFS&REQUEST=GetFeature&VERSION=1.0.0&outputFormat=csv&CQL_FILTER=(realtime=TRUE)"
+        elif realtime.lower() == "no":
+            url = "http://geoserver-123.aodn.org.au/geoserver/ows?typeName=moorings_all_map&SERVICE=WFS&REQUEST=GetFeature&VERSION=1.0.0&outputFormat=csv&CQL_FILTER=(realtime=FALSE)"
+        else:
+            sys.exit('ERROR: %s is not yes or no' % realtime)
+    else:
+        url = "http://geoserver-123.aodn.org.au/geoserver/ows?typeName=moorings_all_map&SERVICE=WFS&REQUEST=GetFeature&VERSION=1.0.0&outputFormat=csv"
```
mhidas: Why did you remove the filter by `site_code` in the WFS query? Didn't that make it much faster?

ANMN/LTSP/TSaggregator/geoserverCatalog_generic.py
```diff
@@ -0,0 +1,112 @@
...
+            sys.exit('ERROR: invalid start date.')
+
+    if timeend:
+        try:
+            criteria_all = criteria_all & (pd.to_datetime(df.time_coverage_start) <=  datetime.strptime(timeend, '%Y-%m-%d'))
+        except ValueError:
+            sys.exit('ERROR: invalid end date.')
+
+    if outFileList:
+        df.url[criteria_all].to_csv(outFileList, index=False)
```
mhidas: The `get_urls` function should simply return the selected urls as a list (this will be the input for the aggregator function). The "__main__" script at the bottom can save it to a file, if requested. Otherwise I think the default should be to simply print it to stdout.

ANMN/LTSP/TSaggregator/geoserverCatalog_generic.py
```diff
@@ -0,0 +1,112 @@
...
+        url = "http://geoserver-123.aodn.org.au/geoserver/ows?typeName=moorings_all_map&SERVICE=WFS&REQUEST=GetFeature&VERSION=1.0.0&outputFormat=csv"
+
+
+    df = pd.read_csv(url)
+    criteria_all = df.url != None
+
+    ## get possible values
+    separator = ', '
+    varnames_all = set(separator.join(list(df.variables)).split(', '))
+    varnames_all = set(str(list(df.variables)).split(", "))
```
mhidas: Do you need both these lines?
Also, you only use  `varnames_all` if `varname` is specified, so why not put it in that `if` clause?

ANMN/LTSP/TSaggregator/geoserverCatalog_generic.py
diodon: Yes, using site_code in the WFS query makes it is a little bit faster.  But I want to check for the validity of the specified site_code when I construct the filtering criteria.

ANMN/LTSP/TSaggregator/geoserverCatalog_generic.py
```diff
@@ -0,0 +1,112 @@
...
+    site_all = list(df.site_code.str.upper().unique())
+    featuretype_all = list(df.feature_type. str.lower().unique())
+    fileversion_all = list(df.file_version.unique())
+    mode_all = ['realtime', 'delayed', 'all']
```
mhidas: Same  with all  of  these `_all` variables. No point working them out if they're not used. Also,
* `site_code` is case sensitive, so don't use `.upper()` for `site_all`
* `feature_type.lower()` should be "timeseries", "profile", or "timeseriesprofile". You don't need to work that out from 46k rows every time.
* `file_version` will be 0, 1, or 2. 
* `mode_all` is not used anywhere, so can be removed.

ANMN/LTSP/TSaggregator/geoserverCatalog_generic.py
diodon: Yes. i want to check that the input varname is a valid one. I need to convert the `variables` string into a list of unique words.

Moved to inside the if clause.

mhidas:
@diodon Sorry, a few more suggestions and questions.

ANMN/LTSP/TSaggregator/geoserverCatalog_generic.py
diodon: fixed

diodon:
no worries @mhidas . Your comments make the code better. Thanks!

ANMN/LTSP/TSaggregator/test/geoserverCatalog.py
diodon: good point. function name changed as suggested.

ANMN/LTSP/TSaggregator/geoserverCatalog_generic.py
diodon: saving the result in a file was now in  __main__. Argument `-outfileName` removed from the function declaration

ANMN/LTSP/TSaggregator/geoserverCatalog_generic.py
mhidas: Ok, but you're making two consecutive assignmets to `varnames_all`, so the second overrides the first. I think you should drop the second line and keep the first.

ANMN/LTSP/TSaggregator/geoserverCatalog.py
```diff
@@ -0,0 +1,99 @@
...
+            sys.exit('ERROR: %s is not in the site_code list' % site)
+
+    if featuretype:
+        #if featuretype in featuretype_all:
+        if featuretype in ["timeseries", "profile", "timeseriesprofile"]:
+            criteria_all = criteria_all & (df.feature_type.str.lower() == featuretype.lower())
+        else:
+            sys.exit('ERROR: %s is not in the feature_type list' % featuretype)
+
+    if fileversion:
```
mhidas: This might not work if the requested file version is 0 (zero), because it is interpreted as False. Instead you should use
```python
if fileversion is not None:
```

diodon: fixed

ANMN/LTSP/TSaggregator/geoserverCatalog.py
```diff
@@ -0,0 +1,97 @@
...
+        if featuretype in ["timeseries", "profile", "timeseriesprofile"]:
+            criteria_all = criteria_all & (df.feature_type.str.lower() == featuretype.lower())
+        else:
+            sys.exit('ERROR: %s is not in the feature_type list' % featuretype)
+
+    if fileversion is not None:
+        if fileversion in [0, 1, 2]:
+            criteria_all = criteria_all & (df.file_version == fileversion)
+        else:
+            sys.exit('ERROR: %s is not in the fileversion list' % featuretype)
```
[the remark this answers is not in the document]
diodon: fixed. `sys.exit` replaced by raise `ValueError`

PR #939, opened 2019-06-17
Geoserver catalog

diodon:


diodon:
added new functionalities to `get_moorings_urls` fucntion:
- Filter by `data_category`
- Filter out the url list by a regex

TSaggregator:
- catch errors when a variable is missing form an input file
- global metadata is generated from a general json file and specific metadata as a concatenated string from input files

diodon:
TSaggr2 is the aggregator function using xarray+pandas. It uses:
- json config file for the global attrs (TSagg_globalmetadata.json)
- json config file for tha variable attrs (TSagg_variableAttributes.json)
- json config file for the input parameter of the function (TSaggr_config.json)

It depends on the geoserverCatalog::get_mooring_url to harvest the urls form the geoserver.

It runs 3x faster and the code is 2x shorter and cleaner. The string variables area `dtype=str` not a char array. This is not CF compliant, but it is easier to handle. It is also very easy to change back to a char array if needed.

ocehugo:
I'm not reviewing frequently, so disregard any possible conflict comment. I think there is room for improvement :)

1. Is this python2 or python3? I can see some "ordered dicts calls" - if you mean to sort by insertion order, python3 is already sorted this way automatically. In fact, do you really need sorted stuff? I don't think so.

2. Why yourr set_globalatt/set_variableatt is loading files and returning structures?  I would load the structure and change it in place - better design and alow re-use of functions.

3. Why also they have different arguments? I would assume from their names that their inputs are the same. Also, why you need two arguments (nc and varname) if you only use nc[varame] and nc[varname+'_quality control']? Small functions for small tasks - I think you just need to compose a better function.

I'm assuming you are taking the previous file as a template. Explicit is better than implicit. I would avoid that and pick the attributes on a specific pre-defined list.

4. Your main_aggregator have implicit options (MainDF, AuxDF,etc). Better collect all of the information required within a single dictionary and setup this as an argument. Putting then into a schema is also a good idea since the scope can grow. This will allow flexibility on the call and even to create a config for each aggregator.

5. The main_aggregator func is too big and need more scoped/semantic actions. Example: You need to load all the files first. Then loop through them. Xarray is smart enough to not clog your memory - lazy loading. When everything is "loaded", you can go check for stuff. Now the aggregation don't need to run to check for errors. You can code tests and test all the structure before commiting to the actual aggregation - a-priori checking. For example:

```
    ....
    all_xr_objs = load_all_files(list_of_files)
    check_agg_objs(al_xr_objs) # > raise/validate error/conditions
    xr_agg = do_agg(all_xr_objs)
    return xr_agg

def write_agg(xr_agg,**opts):
     ...
```

6. Your comments are too long and could be actual tests instead. If your comments deserve more than one line, then it need a proper function. 

7. If you use the pytest library you don't need all the luggage from the unittest library. Your tests comments can be actual tests instead of just noise :)

I didn't dig more since the above is already a lot of food for thought

mhidas:
This PR is now obsolete, isn't it? I think it was superseded by #942. Can we close it?

lbesnard:
@mhidas can this be closed?

PR #940, opened 2019-06-26
Python3 conversion

lbesnard:
NOT TO MERGE

In order to test the following code change work well with python 3, the following commands should run successfully:
 * ```ANMN/NRS_AIMS/REALTIME/anmn_nrs_aims.py```
 * ```FAIMMS/REALTIME/faimms.py```
 * ```SOOP/SOOP_TRV/soop_trv.py```
 * ```SOOP/SOOP_ASF_SST_XBT/SOOP_XBT_NRT.py```
 * ```SOOP/SOOP_ASF_SST_XBT/SOOP_BOM_ASF_SST.py```
 * ```SRS/SRS_OC_LJCO_AERONET/download_aeronet.py /tmp```

 * AUV has been modified as well but untested ```AUV/auv_viewer_processing/auv_processing.py``` since this has some weird gdal dependency installed via chef. So this could be more complicated to deal with. 

There are a bunch of other python code under ANMN, ABOS, AODN and AATAMS which are still using pipeline v1. They need to be converted soon to use pipeline v2.

The following codes are running on personnal machines, or for other users. And don't really need to be updated (as yet):
 * SOOP/SOOP_XBT/DELAYED
 * SRS/SRS_OC_BODBAW

This code was written by Ankit, and should be ported I guess to python 3.5:
 * bin/geonetwork_broken_link_checker.py

@craigrose FYO

### DIFF
--- a/ANMN/NRS_AIMS/REALTIME/dest_path.py
+++ b/ANMN/NRS_AIMS/REALTIME/dest_path.py
@@ -1,99 +1,31 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3.5
 # -*- coding: utf-8 -*-
 """
 Returns the relative path of a ANMN NRS NetCDF file
 author Laurent Besnard, laurent.besnard@utas.edu.au
 """
 
-import os
-import re
-import sys
 
 from netCDF4 import Dataset
-
-
-def get_main_anmn_nrs_var(netcdf_file_path):
-    netcdf_file_obj = Dataset(netcdf_file_path, mode='r')
-    variables       = netcdf_file_obj.variables.keys()
-    netcdf_file_obj.close()
-
-    del variables[variables.index('TIME')]
-    del variables[variables.index('LATITUDE')]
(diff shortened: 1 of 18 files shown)


ANMN/NRS_AIMS/REALTIME/dest_path.py
```diff
@@ -1,99 +1,31 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3.5
```
craigrose: Unless it is necessary should we just use the default python version?

lbesnard: this was a suggestion from @lwgordonimos to do it this way if I understood correctly what the man meant at that time

lwgordonimos: Doing it like these means scripts can be ported over one by one, rather than doing a big bang approach and updating the system interpreter for *everything*.

Also, doesn't need to be specific version, just `/usr/bin/env python3` is fine.

PR #941, opened 2019-06-27
RM old/unused MATLAB/Python code

lbesnard:


plant-image/plant-image.py
```diff
@@ -1,106 +0,0 @@
...
-import warnings
-from datetime import datetime, timedelta
-
-import numpy as np
-from netCDF4 import Dataset
-from PIL import Image
-from scipy import misc
-
-# wget https://imos-data.s3-ap-southeast-2.amazonaws.com/IMOS/SRS/SST/ghrsst/L3U/n19/2015/20151231145731-ABOM-L3U_GHRSST-SSTskin-AVHRR19_D-Des.nc
-# ./plant-image.py 20151231145731-ABOM-L3U_GHRSST-SSTskin-AVHRR19_D-Des.nc gone-skiing-srs.png 19500101000000-ABOM-L3U_GHRSST-SSTskin-AVHRR19_D-Des.nc
```
mhidas: We did remove this image from the portal, right??

lbesnard: yes

lbesnard:
> My only thought is maybe we should leave the temperature gridded code until we have actually removed/replaced that collection with the new products. For now we are still providing files to the public that were generated with this code.

ha ok,  I thought you were keen for that. Have we actually ever used that code ?? I know Bene's code was used, but I don't think my version ever got used

bpasquer:
Just noticed MARVL related code in the SRS directory. I think we should keep it for reference or in case we  decide to update the dataset in the future.
Maybe we can move it to a newly created MARVL directory?

lbesnard:
> just noticed MARVL related code in the SRS directory
Well apparently I'm the author of this code. I don't even know what it's supposed to be doing. I think it would be keeping rubish for nothing (note that it would still be available in the history)

bpasquer:
@lbesnard  ok then we can delete it then

lbesnard:
re-added temp-regridded @mhidas @bpasquer . Feel free to merge

PR #954, opened 2019-10-08
Python3 conversion2

lbesnard:
Similar to https://github.com/aodn/data-services/pull/940 (which should be closed). However, this PR should be compatible with both python2 and 3 (with the exception of AUV as rather untested). 
Code cleaning (indentation ...) across the code was not done to help to review the changes


lwgordonimos:
@lbesnard is this ready for RC?

lbesnard:
yes it is

PR #958, opened 2019-11-11
feat(decomission): ANMN_PA watcher removal

ocehugo:
See also https://github.com/aodn/chef-private/pull/3395


mhidas:
You might as well remove the handler script (https://github.com/aodn/data-services/blob/master/ANMN/acoustic/incoming_handler.sh) in this PR too, as it is now obsolete. (Leave the other code in that directory, as it is still needed).

ocehugo:
@mhidas,

I just renamed the incoming_handler so we don't need to lose the code if we ever need to re-run.

mhidas:
We won't lose the code because it's in github!

ocehugo:
done

mhidas:
@lwgordonimos When you have a minute, could you please use your super-powers to shut down the ANMN_PA pipeline on 10-aws-syd (and 4-nec-hob)?

PR #960, opened 2019-11-12
Remove obsolete Python code

mhidas:
This PR removes Python code, bash scripts and other files that are no longer needed. This should help simplify the task of migrating to Python 3.


mhidas:
@lwgordonimos See latest commit (https://github.com/aodn/data-services/pull/960/commits/3a63c2b9383032a268176c83418233995bc25ccb) for packages I _think_ we can remove from the requirements list.

mhidas:
I have rebased onto master to include #954.

If I unpin all the package versions in requirements.txt, I can successfully install a virtualenv with python 3.6 (otherwise I couldn't).

I still get this test failing:
```
#############################
Executing: 'lib/test/python/test_generate_netcdf_att.py'
#############################
Traceback (most recent call last):
  File "lib/test/python/test_generate_netcdf_att.py", line 17, in <module>
    from generate_netcdf_att import *
  File "/sw/chef/src/data-services/lib/python/generate_netcdf_att.py", line 47, in <module>
    from ConfigParser import SafeConfigParser
ModuleNotFoundError: No module named 'ConfigParser'
```
Apparently it's been renamed to `configparser` in Python 3 (https://docs.python.org/3.6/library/configparser.html)

These two are also failing:
lib/python/test_file_classifier.py
ABOS/ASFS/test_dest_path.py
... but they will soon be obsolete so I won't bother trying to fix them.

mhidas:
I think this could be merged, just to clean up a bit.

The test `lib/test/python/test_generate_netcdf_att.py` is failing in Python 3.6, but Jenkins is still running the tests in Python 2.7, so it won't prevent the code from being deployed to prod.

requirements.txt
```diff
@@ -1,16 +1,10 @@
 GitPython==2.0.6
 OWSLib
-PyYAML
-cf_units
-dotenv
```
lbesnard: I need this! lib/python/aims_realtime_util.py

mhidas: Oops.. I did some grepping of all python code to find what packages were being imported, but somehow I missed that one. I'll add it back in.

PR #963, opened 2019-11-18
Remove functools32 due to incompatibility with python3

anguss00:


lbesnard:
Can't we have a requirement file per python version?

anguss00:
Yeah we could, I am installing the package explicitly in chef for python2. @lwgordonimos thoughts?

anguss00:
Fixed @lbesnard @lwgordonimos

PR #974, opened 2019-12-04
AATAMS DM - replace old lftp bash code with python creating proper ma…

lbesnard:
…nifest file for pipeline 2


lbesnard:
this is needed so the pipeline v2 of aatams dm can work properly with manifest files

PR #978, opened 2020-01-14
Updates to SOOP XBT conversion code

BecCowley:
Add some items to the xbt_config file, including fixing information and typos.
Also update python code to allow for import of files with more than one profile.


ocehugo:
Hi Bec,

I'mpoking @ibesnard since the code above may be backward incompatible (old xbt streams may break with new array access patterns).

lbesnard:
@ocehugo , yes I'm taking this one over. No need to worry, this code is not used in any of our infrastructure

PR #979, opened 2020-01-16
cronjobs to explicitely use python3

lbesnard:


lib/python/IMOSnetCDF.py
```diff
@@ -104,12 +104,12 @@ def close(self):
         self.date_created = time.strftime('%Y-%m-%dT%H:%M:%SZ', time.gmtime())
         if not DEBUG: self.deleteEmptyAttributes()
         self._F.close()
-        if self.__dict__.has_key('tmpFile'):
+        if 'tmpFile' in self.__dict__:
             # rename to desired filename and set permissions
             move(self.tmpFile, self.filename)
-            os.chmod(self.filename, 0644)
+            os.chmod(self.filename, 0o644)
```
bpasquer: pls correct the typo in the mode

bpasquer: sorry my bad, didn't know it is the way to code the chmod .

lib/python/IMOSnetCDF.py
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#! /usr/bin/env python3
```
mhidas: Actually, this is a module, so doesn't really need this line.

lbesnard: absolutely. I removed most shebangs from modules. I must have not think for that one

PR #982, opened 2020-01-23
AATAMS NRT - remove cron job code

lbesnard:
AATAMS/aatams_sattag_metadata.csv is also removed and doesn't seem to be used by anything else


lbesnard:
@lwgordonimos can you please remove the celery bit? this was the last pipeline v1.

The only thing left using v1 is the old async-upload

PR #984, opened 2020-01-23
AATAMS - readd aatams sattag metadata

lbesnard:


bpasquer:
@evacougnon confirmed that the file is not harvested by aatams DM harvester
so closing the PR as  we don't need to restore this file

PR #987, opened 2020-03-03
Checker test updates

mhidas:
I have tried to simplify the config file and output JSON a bit. It may not seem like much, but I think it will be helpful when we add (potentially many) more test cases.

I have also made some tweaks to the code that I think make sense.

### DIFF
--- a/lib/python/compliance_checker_test.py
+++ b/lib/python/compliance_checker_test.py
@@ -130,15 +130,15 @@ def run_test_type_netcdf(test_type, collection_info, tempfile_nc_path):
                 os.makedirs(error_results_path)
 
             # adding a failure key/value in the dictionary output
-            collection_tests_results.setdefault('{test}_failure_filename'.format(test=test), []).append(
-                os.path.join('error_results', err_filename))
+            collection_tests_results['{test}_failure_filename'.format(test=test)] = os.path.join('error_results',
+                                                                                                 err_filename)
 
             os.rename(keep_outfile_path, os.path.join(OUTPUT_DIR, error_results_path, err_filename))  # save file when a test has an error
         else:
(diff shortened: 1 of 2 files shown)


lib/python/compliance_checker_test.py
```diff
@@ -157,20 +157,16 @@ def run_test_all_collection(compliance_config):
         print("Running test suite for: {collection}".format(collection=collection))
 
         for sub_collection in compliance_config[collection].items():
-            try:
```
lbesnard: why removing the try except?

lib/python/compliance_checker_test.py
```diff
@@ -162,8 +160,7 @@ def run_test_all_collection(compliance_config):
 
             # running checks
             for param_results_att in ['check_success_tests', 'check_fail_tests' ]:
-                sub_collection_tests_results = run_test_type_netcdf(param_results_att, sub_collection,
-                                                                    tempfile_nc_path)
+                sub_collection_tests_results = run_test_type_netcdf(param_results_att, info, tempfile_nc_path)
```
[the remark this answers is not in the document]
mhidas: You were passing `sub_collection` to this function, but the only thing it did with it was call `netcdf_tests_info(sub_collection)`. This has already been done here (line 160), and the result stored in `info`, so I thought it was more straightforward for `run_test_type_netcdf()` to accept `info` directly.

lib/python/compliance_checker_test.py
mhidas: Because it has no effect. The only thing it was doing with the caught exception is raising it. That's equivalent to not catching it in the first place, isn't it?

mhidas:
Thanks @lbesnard !

PR #996, opened 2020-06-15
Add more moorings test cases to checker test suite

mhidas:


lbesnard:
@mhidas I don't understand why there is a separate file for moorings. There are some anmn stuffs in both config files now.
Also I really don't care, more up to you, but you could update the abos acronyme

mhidas:
> I don't understand why there is a separate file for moorings.

Just while I was testing them it was helpful to run just the moorings files, so I created a separate config file. Then I just copied all its contents into the main one, which is still the default if you don't specify one on the command line. If you think it's confusing to have two, we can remove the moorings one..

I'm not going to jump the gun on ABOS renaming. We'll do it here when we're doing it everywhere.

lbesnard:
I ran the mooring config file and got some unexpected failures. To me unexpected failures should only happen when we're testing with a new version of the checker:
```
Unexpected failures:
IMOS_ABOS-SOTS_REFOBGTPCS_20090922_Pulse_FV02_Pulse-6-2009-Gridded-Data_END-20100325_C-20190522.nc_cc:4.1.1_imos-plugin:1.4.1_cf:1.6_error_results.txt
IMOS_ANMN-NRS_TZ_20181213_NRSROT_FV01_TEMP-aggregated-timeseries_END-20190523_C-20200723.nc_cc:4.1.1_imos-plugin:1.4.1_cf:1.6_error_results.txt
IMOS_ANMN-NRS_TZ_20181213_NRSROT_FV01_TEMP-aggregated-timeseries_END-20190523_C-20200723.nc_cc:4.1.1_imos-plugin:1.4.1_imos:1.4_error_results.txt

```

mhidas:
> To me unexpected failures should only happen when we're testing with a new version of the checker

Yeah, I know it's a bit confusing, I had to think about it for a while, but I still think this is ok. Those files _should_ pass the checks because the _are_ compliant. So in the configuration file I put them under "check_success_tests" because, quite literally, I would expect them to pass. The reason they are failing is due to bugs in the checker, so in that sense they are unexpected.
Does that make sense?

Anyway, that message printed at the end is just meant to be a helpful summary, and doesn't affect the other outputs. To get the full picture, we will still need to compare the `compliance_checker_results_*.json` files and individual `error_results.txt` files for different versions of the checker.

mhidas:
To put it another way (https://github.com/aodn/data-services/pull/996#issuecomment-662849666), if having an entry in "check_success_tests" means "I know this file passes this check with the version of the checker we currently use", then we have to update the config every time we update the checker to a new version that has either fixed some bugs or introduced others.

Instead, if "check_success_tests" means "I know this file is compliant", that's independent of the version, and if that file actually fails the check, for any version, then something is wrong with that version. Doesn't that make more sense?

Happy to discuss if you disagree. Otherwise please merge.

lbesnard:
no I think you're right, it makes sense. those files should pass. However what is confusing is having anmn config in both config files. I think it's just easier having all in one

mhidas:
Ok. I have removed the separate moorings config file.

PR #998, opened 2020-06-18
CC test suite - add SRS SAR + Scatterometry

lbesnard:


lib/python/compliance_checker_imos_files_config.json
```diff
@@ -239,5 +239,32 @@
       ],
       "check_fail_tests": []
     }
+  },
+  "srs_surface_waves_scatterometry": {
+    "file_url": "https://s3-ap-southeast-2.amazonaws.com/imos-data/IMOS/SRS/Surface-Waves/Wind-Scatterometry-DM00/ERS-1/000N_000E/IMOS_SRS-Surface-Waves_M_Wind-ERS-1_FV02_000N-000E-DM00.nc",
```
mhidas, line 244: Will these files likely get replaced by a new version with a different name?
If so, you might want to make a separate copy in [IMOS/eMII/test_files/](http://imos-data.s3-website-ap-southeast-2.amazonaws.com/?prefix=IMOS/eMII/test_files/) to avoid having to update this config.

lbesnard: No they shouldn't. but even if they were, they're should pass the same checker version. I think that once we have a good set of files (for all of them) we're happy with, we can then decide where to have them all, and add a commit to update the file_url for all of them. A quick ```sed``` should do it

PR #1002, opened 2020-07-21
Update unit view with new unit for aatams

bpasquer:


bpasquer:
Not that the "id" field is now irrelevant. The new vocab uses a mix of number and string. The "id" added here was chosen randomly.

PR #1011, opened 2020-08-13
Updates to improve XBT netcdf format

BecCowley:
Changed the *ADJUSTED variables to be TEMP and DEPTH and the raw data is now in TEMP_RAW and DEPTH_RAW. This is contrary to the Argo method, but is more logical from a user point of view.
Also added the probe type name to the fall rate statement and did some deletion of duplicated QC flags.


lbesnard:
@BecCowley It would be great if you could also edit accordingly the unittests in the ./test_xbt_dm_imos_conversion.py file as this would give us confidence in current and future changes

BecCowley:
@lbesnard, apologies if I did it wrong. I am a bit ignorant about how best to do the tests and to use the unittests. I will have another look, but feel free to give me lots of advice!

lbesnard:
no worries at all! The unittests are defined here:
https://github.com/aodn/data-services/blob/master/SOOP/SOOP_XBT/DELAYED/test_xbt_dm_imos_conversion.py

The framework is already there, so it shouldn't be to hard to modify accordingly.
The idea is when some work is done on the main script, to run in your terminal/IDE (I use pycharm to ease debugging) ```test_xbt_dm_imos_conversion.py```  to check that the main script still behaves as expected, in order to spot un-wanted changes. In your case you have changed quite a lot of things, including parameter names, such as https://github.com/aodn/data-services/blob/master/SOOP/SOOP_XBT/DELAYED/test_xbt_dm_imos_conversion.py#L196 

So the unittests need to reflect those kind of changes. It may sound tedious but it adds so much more confidence in amending the/any code that it's really worth it. 
You can pretty much test anything you want within unittests, but it's best to find the right balance.

BecCowley:
@lbesnard , thanks. I will review next week and hopefully learn something in the process!

BecCowley:
@lbesnard, I have made some udpates to the files (as you can probably see). Thanks for educating me about unittests, I can see the value.
Let me know if I need to do anything to help merge these changes.

SOOP/SOOP_XBT/DELAYED/xbt_dm_imos_conversion.py
```diff
@@ -407,15 +407,17 @@ def raw_for_ed_path(netcdf_file_path):
...
+    
+    #decide what prefix is required
     names = read_section_from_xbt_config('VARIOUS')
     str = names['FILENAME']
     if str == 'Cruise_ID':
         str = gatts['XBT_cruise_ID']
-
-    filename = '%s-XBT_T_%s_%s_FV01_ID-%s' % (str, data['TIME'].strftime('%Y%m%dT%H%M%SZ'), gatts['XBT_line'], gatts['XBT_uniqueid'])
-
-#    if data['TIME'] > datetime(2008, 0o1, 0o1):
```
lbesnard: regarding this part, not that I particularly care and remember exactly why the filename was different after this arbitrary date (I would have to chat with Seb), but do you want in the future to reprocess the whole dataset?

BecCowley: I think the name difference is because funding for IMOS was only after 2008? Hence the IMOS label on the later data and the XBT-SOOP on the earlier data.
I don't know if we need to keep the different names or not - need to check with Seb, I think.

lbesnard:
@BecCowley I'll merge and chat to Seb

PR #1012, opened 2020-08-19
Move cleanup of soop_trv temp file to a finally block

lwgordonimos:


lbesnard:
@lwgordonimos I reviewed and would like to merge. happy?

lwgordonimos:
Yeah if it works! I didn't test it at all, so if you've tested it and are happy, I'm happy.

PR #1015, opened 2020-09-02
 add Buoys metadata file to data-services repo

bpasquer:


bpasquer:
This PR is an improvement to the current pipeline for aodn_wave_nrt : 
Currently, talend harvest this file stored in the WIP directory. The file was not version-controlled (i know it was baaaaad)
So this is fixing it

PR #1019, opened 2020-09-21
python code to change ABOS attributes and filenames

ana-berger:


DWM/fix_ABOSncFiles_to_DWM.py
```diff
@@ -0,0 +1,75 @@
...
+
+def fix_abos_attributes(ncobj: nc.Dataset, namelist: List[str], old: str = 'ABOS', new: str = 'DWM'):
+    for attname in namelist:
+        attvalue = getattr(ncobj, attname)
+        new_attvalue = replace_abos_str(attvalue, old, new)
+        setattr(ncobj, attname, new_attvalue)
+
+
+def get_netcdf_files(folder: str):
+    return glob(os.path.join(folder, '**/*.nc'), recursive=True)
```
lbesnard, line 46: this is a very slow approach using the s3 mount. works, but would be better to list the files using WFS

DWM/fix_ABOSncFiles_to_DWM.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def rename_abos_file(old_filename: str):
+    if 'ABOS' not in old_filename:
+        warnings.warn(f"Rename requested for {old_filename}")
+        return
+    else:
+        new_filename = old_filename.replace("ABOS", "DWM")
+        print(f"Renaming {old_filename} -> {new_filename}")
+        os.rename(old_filename, new_filename)
```
lbesnard, line 70: you can't do this. I think you would need to have a proper chat with @mhidas on the pipeline infrastructure. 
you need to copy the old ABOS files to a tmp folder, modify them into DWM and then push them to the new incoming dir. Then delete the original files with po_s3_del bash function available on 4-nec/10-aws.

This script doesn't have to run on prod. it could be done on your local machine, and then you would have to scp the new files to 10-aws

DWM/fix_ABOSncFiles_to_DWM.py
```diff
@@ -0,0 +1,75 @@
...
+from glob import glob
+
+
+# def test_replace_abos_str():
+#     string = 'my_test_XXX'
+#     old = 'XXX'
+#     new = 'YYY'
+#     expected = 'my_test_YYY'
+#     result = replace_abos_str(string, old, new)
+#     assert(result == expected)
```
lbesnard, line 17: no need to have this

lbesnard (changes requested):
not that it is a complicated script, but more comments would be good. also a proper intro at the top would explain what this is about in 5 years time

DWM/fix_ABOSncFiles_to_DWM.py
ana-berger: it will be very slow to copy all the files, but if this is the only way of doing then we do that. But just curious, why it cannot be done in prod? I could make a copy of them in prod and then substitute the old files after, but being there already would be much faster

lbesnard: you just don't have the choice. the files are on a s3 bucket. it doesn't work as a standard filesystem. and even if it did, no PO's would have writing access anyway. The only way is to use the incoming dir from the pipeline

mhidas (changes requested):
Welcome to the wonderful world of code reviews!
Just a couple of things to fix, otherwise it looks good, and will do what we need to do.

DWM/fix_ABOSncFiles_to_DWM.py
ocehugo: Yes, the problem here is that ana changed the "tmp" folder to the s3. The script is supposed to work on a temporary folder since we are doing attribute replacement.

ana-berger: I didn't mention but that folder in 'dataDIR' will change, that is only a temporary address. I didn't even try to run in that path, I run it only locally with some test files

DWM/fix_ABOSncFiles_to_DWM.py
```diff
@@ -0,0 +1,75 @@
...
+        warnings.warn(f"Rename requested for {old_filename}")
+        return
+    else:
+        new_filename = old_filename.replace("ABOS", "DWM")
+        print(f"Renaming {old_filename} -> {new_filename}")
+        os.rename(old_filename, new_filename)
+
+
+if __name__ == '__main__':
+    dataDIR1 = '/mnt/imos-test-data/IMOS/ABOS/'
```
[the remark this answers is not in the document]
ana-berger: yes sorry, I should have added the comment that the folder would change, this was just an example in case it was running in rc, which I was trying to do earlier today when I shared the code. But originally it was run in my local path with some example files

mhidas:
My only other comment is about the name of your script (`fix_ABOSncFiles_to_DWM.py`). This really just a personal preference, but this one mixes every possible naming convention (underscores, lower/uppercase, initial capitals). For Python scripts we generally just use all lower case and underscores between words (see [here](https://github.com/aodn/data-services/issues/283) for a discussion on our Python coding conventions).

In this case it's just a stand-alone script, so not a big deal, but in case you're curious, I would name it something like `update_abos_attributes_to_dwm.py`

DWM/fix_ABOSncFiles_to_DWM.py
```diff
@@ -0,0 +1,75 @@
...
+    return glob(os.path.join(folder, '**/*.nc'), recursive=True)
+
+
+def fix_abos_files(folder: str):
+    files = get_netcdf_files(folder)
+
+    for file in files:
+        ncobj = nc.Dataset(file, 'a')
+        abos_attributes = find_att_with_str(ncobj, 'ABOS')
+        abos_blue_attributes = find_att_with_str(ncobj, 'Australia Bluewater Observing System')
```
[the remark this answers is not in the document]
ana-berger: it is actually as "Australia" in the files

ana-berger:
I considered all the comments and did the appropriate changes:
- added explanation comments and header
- changed files path to the `sys.argv[1]` as suggested
- all files I tested have Australia instead of Australian (... Bluewater Obs System)
- deleted one unnecessary function and comments 
- added conditional to exclude the **"history"** attribute that should not be changed
- renamed the script filename

I think that was all!

ana-berger:
> Welcome to the wonderful world of code reviews!
> Just a couple of things to fix, otherwise it looks good, and will do what we need to do.

yayyy!!

DWM/update_abos_attributes_to_dwm.py
```diff
@@ -1,51 +1,43 @@
 #!/usr/bin/env python3
 """Rename ABOS files and attributes"""
+# As part of the ABOS renaming Odyssey, this script identifies the attributes that contains the words "ABOS" and
+# "Bluewater" (as art of the Australian Bluewater Observation System), changes them for "DWM"/"Deep Water Moorings"
+# and rename the files. The attribute 'history' is excluded from the search as we decide that attribute did not need
+# to be changed.
```
mhidas, line 6: Maybe we don't need to refer to the "ABOS renaming odyssey" in a public repo...
Also, you can make this paragraph part of the "docstring" for the script, by including it in the triple-quoted text starting on line 2:
```python
"""Rename ABOS files and attributes

This script identifies the attributes...
"""
```
Also a couple of typos in there (contain~~s~~, art -> part)

DWM/update_abos_attributes_to_dwm.py
```diff
@@ -1,51 +1,43 @@
...
-# def test_replace_abos_str():
-#     string = 'my_test_XXX'
-#     old = 'XXX'
-#     new = 'YYY'
-#     expected = 'my_test_YYY'
-#     result = replace_abos_str(string, old, new)
-#     assert(result == expected)
-
+# search in the attributes which one has 'ABOS' in it, excluding the 'history' attribute (we do not need to change this)
 def find_att_with_str(ncobj: nc.Dataset, string: str = 'ABOS'):
```
mhidas: Instead of comments above each function, these descriptions can also become [docstrings](https://www.datacamp.com/community/tutorials/docstrings-python) for each function, e.g.
```python
def find_att_with_str(ncobj: nc.Dataset, string: str):
    """Search in the attributes which one has 'ABOS' in it, excluding the 'history' attribute """
```

DWM/update_abos_attributes_to_dwm.py
```diff
@@ -1,51 +1,43 @@
...
-    # for k,v in global_attr_dict:
-    #    got_requested_string = type(v) is str and string in v
-    #    if got_requested_string:
-    #        attrs.append(k)
-
     for attname in ncobj.ncattrs():
         attvalue = getattr(ncobj, attname)
         got_requested_string = type(attvalue) is str and string in attvalue
-        if got_requested_string:
+        if got_requested_string and attname != 'history':
```
mhidas, line 23: We should also exclude the attribute "toolbox_input_file"

mhidas (commented):
Getting there!

DWM/fix_ABOSncFiles_to_DWM.py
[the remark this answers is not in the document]
mhidas: Ah, you're right. Good spotting!

mhidas (approved):

PR #1020, opened 2020-09-23
improve-and-add-new-coding-to-save-log_file

ana-berger:


DWM/update_abos_attributes_to_dwm.py
```diff
@@ -38,34 +37,42 @@ def get_netcdf_files(folder: str):
...
     for file in files:
         ncobj = nc.Dataset(file, 'a')
         abos_attributes = find_att_with_str(ncobj, 'ABOS')
         abos_blue_attributes = find_att_with_str(ncobj, 'Australia Bluewater Observing System')
         fix_abos_attributes(ncobj, abos_attributes, old='ABOS', new='DWM')
-        fix_abos_attributes(ncobj, abos_blue_attributes, old='Australia Bluewater Observing System',
-                            new='Deep Water Moorings')
+        fix_abos_attributes(ncobj, abos_blue_attributes, old='Australia Bluewater Observing System', new='Deep Water Moorings')
+        for attr_name in abos_attributes:
+            logging.info("New attribute: {}".format(getattr(ncobj, attr_name)))
```
mhidas, line 51: Why not log the name of the updated attribute as well?
```suggestion
            logging.info("New value of {name}: '{value}'".format(name=attr_name, value=getattr(ncobj, attr_name)))
```
And also do the same for `abos_blue_attributes`?

DWM/update_abos_attributes_to_dwm.py
```diff
@@ -38,34 +37,44 @@ def get_netcdf_files(folder: str):
...
     return glob(os.path.join(folder, '**/*.nc'), recursive=True)
 
- 
+
 def fix_abos_files(folder: str):
     """ Call the functions to replace attributes that contain 'ABOS' and/or 'Australia Bluewater Observing System') """
     files = get_netcdf_files(folder)
-
     for file in files:
         ncobj = nc.Dataset(file, 'a')
```
mhidas, line 45: Add a log line here. Something like "Updating file {file}"

ana-berger: I think this is all fixed now

mhidas: :+1:

PR #1022, opened 2020-10-13
add-other-attribute-to-be-fixed

ana-berger:
I found some attributes that have "Australia**n**" Bluewater Obs System, instead of Australia Bluewater Obs System like most of them. So just added another condition

### DIFF
--- a/DWM/update_abos_attributes_to_dwm.py
+++ b/DWM/update_abos_attributes_to_dwm.py
@@ -39,7 +39,7 @@ def get_netcdf_files(folder: str):
 
 
 def fix_abos_files(folder: str):
-    """ Call the functions to replace attributes that contain 'ABOS' and/or 'Australia Bluewater Observing System') """
+    """ Call the functions to replace attributes that contain 'ABOS' and/or 'Australia(n) Bluewater Observing System') """
     files = get_netcdf_files(folder)
     for file in files:
         print("Updating file: {}".format(file))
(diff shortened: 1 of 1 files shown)


DWM/update_abos_attributes_to_dwm.py
```diff
@@ -39,20 +39,24 @@ def get_netcdf_files(folder: str):
...
     for file in files:
         print("Updating file: {}".format(file))
         logging.info("Updating file: {}".format(file))
         ncobj = nc.Dataset(file, 'a')
         abos_attributes = find_att_with_str(ncobj, 'ABOS')
         abos_blue_attributes = find_att_with_str(ncobj, 'Australia Bluewater Observing System')
+        abos_blue2_attributes = find_att_with_str(ncobj, 'Australian Bluewater Observing System')
         fix_abos_attributes(ncobj, abos_attributes, old='ABOS', new='DWM')
         fix_abos_attributes(ncobj, abos_blue_attributes, old='Australia Bluewater Observing System', new='Deep Water Moorings')
+        fix_abos_attributes(ncobj, abos_blue2_attributes, old='Australian Bluewater Observing System', new='Deep Water Moorings')
```
mhidas, line 53: This is totally fine, but starting to get a bit repetitive. If you had more than three cases to deal with, it would be worth generalising it by creating a dictionary of substitutions (old: new) you want to do, and then iterating through it...
```python
    substitutions = {'ABOS': 'DWM',
                     'Australia Bluewater Observing System': 'Deep Water Moorings',
                     'Australian Bluewater Observing System': 'Deep Water Moorings'
                     }
    for file in files:
        # ...
        for old, new in substitutions.items():
            attribute_list = find_att_with_str(ncobj, old)
            fix_abos_attributes(ncobj, attribute_list, old=old, new=new)
```
Also, the logging line for each attribute could be moved into the `fix_abos_attributes` function, so you don't have to explicitly do it here at all.
(And finally, you could also use regular expressions ([`re.search`](https://docs.python.org/3.6/library/re.html#re.search)/`re.sub`) instead of an exact sub-string match to combine the two variants of the full facility name)

mhidas:
As a more general comment, the title of a pull request is just a string and can have spaces in it. There's no need to join all the words with -'s (unlike the name of a _branch_, which cannot contain spaces...)

ana-berger:
ah I think I am not adding a title and then it copies the same name as the branch, that is why is looking like a branch name. But I will start adding a normal title without -'s. Thank you

PR #1032, opened 2021-01-21
Xbt srfc code fixes

BecCowley:
Add new surface codes and adjust reading to limit to the number of surface codes in Nsurfc variable.


BecCowley:
Closing to change name of branch and add more commits. Will resubmit under a new PR.

PR #1044, opened 2021-04-21
feat(ACORN): bash scripts for hfradar reports

ocehugo:
This includes most of my scripts for hf radar
quick & dirty reporting and queries.

Useful for a quick glance at the status of the
error directory, database tables, missing files,
etc.

IMO, the most useful feature is a single command
to verify if files in the error directory
are already included in the database.
It is a bit slow sometimes, but it gets the job done.


ocehugo:
@lbesnard - I just added some extra little fixes - see commit msgs.

lbesnard (approved):

PR #1047, opened 2021-05-03
updates to process BOM DM wave data

ana-berger:


requirements.txt
```diff
@@ -15,11 +15,14 @@ pyexcel-xlsx==0.5.8
 retrying==1.3.3
 scipy==0.14.1; python_version < '3.6'
 scipy==1.4.1; python_version > '3.5'
-six
+six~=1.15.0
```
mhidas, line 18: Did you add these version constraints in the requirements file? Or did PyCharm add them?
Need to be careful with these, as locking in versions can cause problems wherever a python environment is set up for this repository. I suggest only including changes that are really necessary (e.g. if an older or newer version of a package breaks the code here).

mhidas: BTW, we shouldn't need `six` anymore (that was for compatibility while migrating from Python 2 to 3) - but that's an issue separate from this PR.

AODN/AODN-WAVE-DM/BOM-WAVE-DM/bom_wave_library/bom_wave_parser.py
```diff
@@ -78,13 +81,13 @@ def parse_xls_xlsx_bom_wave(filepath):
...
-        is badly written and and some leading and trailing spaces in datetime variable (for example cpso2004.xlsx, cdec2007)
-        Also some original files had a date written as  03/09/2003 05:00:001 <- ":001" !! so they were manually edited
+        is badly written and some leading and trailing spaces in datetime variable (for example cpso2004.xlsx, cdec2007)
+        Also some original files had a date written as 03/09/2003 05:00:001 <- ":001" !! so they were manually edited
         In this case we do the following
         """
 
         if isinstance(df['datetime'].values[0], unicode) or isinstance(df['datetime'].values[0], str):
-            date_format = '%d/%m/%Y %H:%M:%S'
+            date_format = '%d/%m/%Y %H:%M'
```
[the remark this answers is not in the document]
ana-berger: > would be good to retain the date format for data 1998-2017 and make a conditional choice based on the file name or the year?

I tried to fix it, not sure is the best way

requirements.txt
ana-berger: I think all the changes made in requirements were suggested by PyCharm and I accepted them, so I imagine they were necessary. Not sure what to do about it.

AODN/AODN-WAVE-DM/BOM-WAVE-DM/bom_wave_library/bom_wave_parser.py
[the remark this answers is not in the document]
bpasquer: I'm happy with that, does the job

PR #1050, opened 2021-05-20
AODN WAVE DM - migrate code to python 3.8 - add env yaml for easy dep…

lbesnard:
…loy with conda


lbesnard:
Migration of the codes to python 3.8. Check readme mods to see how to use

haven't fully tested, but the codes run. Seems like a lot of things of changed in the back end of DES and DOT, so further mods are required

lbesnard:
@ana-berger probably should be merged as it's an improvement over the previous code

PR #1054, opened 2021-06-24
Create temporary extract directory in system temporary directory

jonescc:


jonescc:
Just checking where environment variables are defined - there's a mismatch for those as well

jonescc:
envirnoment variables are defined in chef-private so this ones good to go

PR #1058, opened 2021-08-04
(Feat) standalone python script to list imos-data objects

lbesnard:
A standalone python script to output object to stdout. Can then easily be used
with po_s3_del
```bash
./s3_ls_prefix.py -h
usage: s3_ls_prefix.py [-h] [-p PREFIX] [-b BUCKET]

optional arguments:
  -h, --help            show this help message and exit
  -p PREFIX, --prefix PREFIX
                        s3 prefix - Example ''IMOS/SOOP'
  -b BUCKET, --bucket BUCKET
                        s3 bucket - default value is 'imos-data'

```

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,3 +1,7 @@
+# An environment file for Project Officers wanted to use Conda rather than
+# virtualenv
+# conda env create --file=environment.yml
+# conda env update --file=environment.yml
 name: data_services_3.6
 channels:
   - conda-forge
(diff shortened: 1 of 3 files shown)


lbesnard:
FYI @aodn/data-ops  can be useful to some of you

leonardolaiolo:
This is just for reference, as still does not work for me. 
These 2 lines can be useful to install AWS correct version. 
```
curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
unzip awscliv2.zip
sudo ./aws/install
```

PR #1060, opened 2021-08-09
(Fix) conda env broken due to pip update semantic change

lbesnard:


lbesnard:
same issue as what we had on python-aodndata repo

PR #1062, opened 2021-08-17
Soop asf meteo flux flag

leonardolaiolo:
Related to https://github.com/aodn/content/issues/450


lbesnard:
@leonardolaiolo why did you remove info related to the qc_scheme_id?

leonardolaiolo:
@lbesnard I did that because there is no qc data, so qu headers where not useful

bpasquer (approved):

PR #1070, opened 2021-09-22
Delete AODN_Keywords_Thesaurus.txt

akashisama:
Hasn't been updated for 7 years so clearly Nat's not using it anymore

Part of https://github.com/aodn/backlog/issues/2518 but useful to get rid of files like this anyway


akashisama:
Confirmed with Nat that this can be deleted

PR #1081, opened 2022-02-15
(RM) Remove pipeline v1 unused functions and unit tests

lbesnard:


lbesnard:
as part of https://github.com/aodn/PO-Backlog/issues/2468,
I removed all the different unused functions from the pipeline V1 which aren't used anymore.

The only ones left are using ```po_s3_del```

I did a thorough check, and checked the pipelines as well as the good working order of the po_s3_del command on the pipeline box.

digorgonzola:
I still get loads of errors when I run these tests in an Ubuntu 20 docker environment with Python 3.8. Not sure if they are significant

lib/common/util.sh
```diff
@@ -4,88 +4,6 @@
...
-
-    return $retval
-}
-export -f _bulk_index_operation
-
-# bulk index files using talend
-# $1 - directory to cd to before running harvester
-# $2 - base to index files with (prefix, such as IMOS/Argo)
-# $3 - file containing list of files to index
-index_files_bulk() {
```
mhidas, line 41: This function, along with `unindex_files_bulk` could occasionally still be useful.
You can get by just calling `$HARVESTER_TRIGGER` directly, but it won't automatically save the log.

mhidas (commented):
Looks ok to me.
The bulk harvesting/unharvesting functions could be kept until there's an alternative that uses the pipeline infrastructure. I just did a quick check in the `pipeline-ubuntu20` box, and they do work under Ubuntu 20.

lib/common/util.sh
lbesnard: I re-added them.

mhidas: :+1: Thanks.

mhidas (approved):

lbesnard:
There was a deprecating error on the ubuntu 20 VM 
```
#############################                                                                                                                                                                                                                                           [101/4951]
#############################
Executing: 'lib/test/python/test_nc_checker_call.py'
#############################
/usr/local/lib/python3.8/dist-packages/tabulate.py:7: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
  from collections import namedtuple, Iterable
```

which can be fixed by upgrading the python system package:
```
sudo pip install tabulate --upgrade
```

digorgonzola:
> udo pip install tabulate --upgrade

This caused me issues in my environment due to a conflict with `compliance-checker` wanting a specific version of `tabulate`.

I've created another branch with some changes to the requirements. Are you able to check? https://github.com/aodn/data-services/pull/1082

PR #1094, opened 2022-07-11
(Fix) Github actions

lbesnard:


lbesnard:
locally tested with docker and https://github.com/nektos/act

digorgonzola (approved):

PR #1099, opened 2022-07-27
(FIX) ARDC_NRT - add retry for BOM wfs

lbesnard:


lbesnard:
do you know about 
```
./setup.py develop
```

so that you can modify the code without have to re-install the package anytime there is a change?

bpasquer (approved):

PR #1102, opened 2022-08-05
Reprocessing historical wave data

ana-berger:


bpasquer:
What should we do about this PR @ana-berger

lbesnard:
FYI, you shouldn't put your .idea folder in this PR . easier if you add it to your .gitignore list

ana-berger:
Well, I have the working version on my machine and the files were already processed using this code. It depends on whether we want the updated version as the master. But not sure it will work the same for everyone because of conda etc.. I had to make adaptations to make it work for me.

bpasquer (approved):

PR #1104, opened 2022-08-16
(Fix) AIMS script update MD5 so that data can be re-downloaded

lbesnard:


ggalibert:
@bpasquer can you please look after this? The changes will need to be deployed to prod and hopefully we should be able to see more recent datasets for SOOP TRV, FAIMMS and NRS YON/DAR. The NRT flow for these dataset has been interrupted since 28 May...

bpasquer (approved):

PR #1113, opened 2022-09-23
Wave improvements

bpasquer:


AODN/AODN-WAVE-NRT/ARDC_API_NRT/ardc_nrt/lib/common/netcdf.py
```diff
@@ -85,7 +85,7 @@ def convert_wave_data_to_netcdf(self, true_dates=False):
         template.add_extent_attributes(time_var='TIME', vert_var=None, lat_var='LATITUDE', lon_var='LONGITUDE')
         template.add_date_created_attribute()
         template.global_attributes.update({
-            'history': "{date_created}: this file was file created on".format(
+            'history': "this file was file created on: {date_created}".format(
```
ggalibert, line 88: This is not following the current IMOS conventions:
> Provides an audit trail for modifications to the original data. It should contain a separate line for each modification, with each line beginning with a timestamp and including user name and modification arguments.

Ex.:
> history = "2014-03-27T23:46:36Z - timeOffsetPP: TIME dimension and time_coverage_start/end gloabal attributes have been appliead the following offset: -10 hours."

bpasquer: I didn't remember/know we had a convention for "history". And I find the statement very confusing. We don't apply an offset to the data since we get it in UTC already

ggalibert: Date should come first, then info about what happened.

ana-berger: We actually don't have 'history' in the templates. It should not be there as far as I know. Maybe I missed something...?

ana-berger (approved):

PR #1115, opened 2022-09-28
(Fix) AIMS scripts update to cf:1.6 instead of cf:latest for compliance checker

lbesnard:


leonardolaiolo:
@lbesnard, just want to double check you want `imos=1.3`? I usually see  `imos=1.4` on other places

lbesnard:
yeap , it matches this: https://github.com/aodn/chef-private/blob/master/data_bags/imos_po_watches/FAIMMS.json#L10

leonardolaiolo (approved):

PR #1118, opened 2022-10-06
AODN WAVE - fix standard name vocab string to trigger appropriate CF check

lbesnard:


AODN/AODN-WAVE-NRT/ARDC_API_NRT/test_ardc_nrt/template_uwa.json
```diff
@@ -163,5 +163,5 @@
   "author_email": "info@aodn.org.au",
   "principal_investigator": "",
   "principal_investigator_email": "",
-  "standard_name_vocabulary": "NetCDF Climate and Forecast (CF) Metadata Convention Standard Name Table 60"
+  "standard_name_vocabulary": "NetCDF Climate and Forecast CF Standard Name Table 60"
```
lbesnard, line 166: @ana-berger said it should be 78

ana-berger: The standard in the template is the exact expression "NetCDF Climate and Forecast (CF) Metadata Convention Standard Name Table 78" (actually 'standard' and not 'Standard'). There were times I run the compliance checker and it was complaining, even about the capital letter. So it had to be that exact expression. I test now and it seems to accept both, but for consistency, we should now keep the same pattern.

lbesnard: Please see link above 
https://github.com/ioos/compliance-checker/blob/2f92a132fa5c6863b37e8b024cb390930aee22b2/compliance_checker/cf/cf.py#L133

which tests for this explicitly:
```python
if 'cf standard name table' in standard_name_vocabulary.lower():
``` 

The line I modified had CF in brackets, so the condition wasn't true. Maybe in your case the checker was running fine because the standard name table on your machine was at a version high enough to check the standard name values , but otherwise it would just fail

bpasquer (approved):

PR #1120, opened 2022-11-02
(Fix) AODN WAVE - use shutil.copy2 instead of move creating issues with celery

lbesnard:


lbesnard:
The "non writable" issue in the handler logs on pipeline-rc were replicable in the pipeline-box which was a start. The issue was with ```shutil.move``` which creates a wrong incron event. Better to do a copy and a delete. 
This issue existed already with other data-services script (SOOP_TRV.py for example)

bpasquer (approved):

PR #1124, opened 2022-11-15
Fix: ANDS vocab seems to have an expired certificate. remove signing of it

lbesnard:


lbesnard:
@BecCowley This is related to the SOOP XBT processing

BecCowley:
@lbesnard thanks, I will merge it when finalised.

atkinsn:
ANDS changed their name to ARDC many years ago. They are just about to stop offering a redirect from their old vocab service name to their new one, and suspect that is why the SSL certificate has not been renewed. My bad, I was aware of this coming up, and I didn't realise the vocab service URL was used here.

New one - https://vocabs.ardc.edu.au/

bpasquer (approved):

PR #1139, opened 2023-01-30
ARDC:PPA buoy metadata

bpasquer:


AODN/AODN-WAVE-NRT/ARDC_API_NRT/ardc_nrt/config/omc/sources_id_metadata.json
```diff
@@ -19,7 +19,7 @@
     "wave_buoy_type": "directional"
   },
   "94ab7906-19f1-4920-a0b3-61e2ee0766c0": {
-    "id": "94ab7906-19f1-4920-a0b3-61e2ee0766c0",
+    "id": "WRB-03",
```
lbesnard, line 22: isn't this breaking the code if the ID not similar to the key?

bpasquer: No it's not breaking anything. In fact this id is only used once in the code for retrieving information only. But anyway, I will change it back for consistency with the other config

lbesnard (approved):

PR #1142, opened 2023-02-16
Reprocessing historical wave data

ana-berger:


requirements.txt
```diff
@@ -1,13 +1,13 @@
 GitPython==2.0.6
 OWSLib
-geopy==1.11.0
```
lbesnard, line 3: you don't need to modify this file, it has it's own requirements.txt file https://github.com/aodn/data-services/blob/master/AODN/AODN-WAVE-DM/requirements.txt

lib/python/generate_netcdf_att.py
```diff
@@ -47,7 +47,7 @@
 
 import numpy as np
 
-from IMOSnetCDF import attributesFromIMOSparametersFile
+from python.IMOSnetCDF import attributesFromIMOSparametersFile
```
lbesnard, line 50: This will break other codes. please revert

AODN/AODN-WAVE-DM/MHL/process_MHLwave_from_txt.py
```diff
@@ -11,13 +11,13 @@
 import pandas
 import numpy as np
 from netCDF4 import Dataset, date2num
-from generate_netcdf_att import generate_netcdf_att
+from python.generate_netcdf_att import generate_netcdf_att
```
lbesnard, line 14: This ```python.``` should be removed

ana-berger: That's the only way it worked on my machine. I am not changing it

lbesnard: on your machine yes maybe, but that's not the way sorry. if this was used on RC or prod, it shouldn't have python.

ana-berger:
This is just for Bene to have the updated code. I am not changing anything as this is the version that worked on my machine. If the working code for reprocessing is to be shared with everyone, that is it. Not doing any work on that anymore

PR #1176, opened 2023-07-25
Wave: update Vic network

bpasquer:


AODN/AODN-WAVE-NRT/ARDC_API_NRT/ardc_nrt/config/sofar/sources_id_metadata.json
```diff
@@ -200,6 +200,30 @@
     "water_depth": 50,
     "water_depth_units": "m",
     "wave_buoy_type": "directional"
+  },
+   "SPOT-1448":
```
evacougnon, line 204: is it normal to have twice the same spotter_id `SPOT-1448`? for a different site_name?

bpasquer: Technically it's got not impact cause these spotters are not processed, but I'll changes it anyway

evacougnon (approved):

PR #1189, opened 2024-01-24
UWA deployment updates

bpasquer:


AODN/AODN-WAVE-NRT/ARDC_API_NRT/ardc_nrt/config/sofar/sources_id_metadata.json
```diff
@@ -352,25 +352,25 @@
   },
   "SPOT-0162":
```
evacougnon, line 353: Shall you update this line too, to have a matching SPOT number?

Same comment for  line 405 `"SPOT-30848C":` should be `"SPOT-31395C"`

evacougnon (approved):

PR #1208, opened 2024-09-06
Delete data associated with NaN lat/long coordinates

xhoenner:
Addresses error described in https://github.com/aodn/content/issues/522

### DIFF
--- a/AODN/AODN-WAVE-NRT/ARDC_API_NRT/ardc_nrt/ardc_sofar_nrt.py
+++ b/AODN/AODN-WAVE-NRT/ARDC_API_NRT/ardc_nrt/ardc_sofar_nrt.py
@@ -67,7 +67,7 @@ def process_wave_source_id(source_id, incoming_path=None):
             return
 
         if data is not None:
-            data = data[data['latitude'].notnull()]
+            data = data.dropna(subset = ['latitude', 'longitude'], how = 'all') # Remove rows associated with NAs for lat/long columns
             template_dirpath = config.conf_dirpath
             process_wave_dataframe(data, source_id, template_dirpath, OUTPUT_PATH, incoming_path)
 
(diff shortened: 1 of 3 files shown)


AODN/AODN-WAVE-NRT/ARDC_API_NRT/ardc_nrt/ardc_sofar_nrt.py
```diff
@@ -67,6 +67,7 @@ def process_wave_source_id(source_id, incoming_path=None):
             return
 
         if data is not None:
+            data = data[data['latitude'].notnull()]
```
bpasquer, line 70: there may be instance when longitude is null  so worth also excluding these values

xhoenner: Thanks @bpasquer. Hopefully addressed your comments via https://github.com/aodn/data-services/pull/1208/commits/06969cc4f401474d857736920e02f077d382d9a0

lbesnard:
could you also fix the 2 github action files in this repo as done in 
https://github.com/aodn/aodn_cloud_optimised/commit/2193da40a0697c861183d422a0b6afe43c42cba8

so that the unittest can run ? 

there have been some changes lately, and i got the same bug

xhoenner:
> could you also fix the 2 github action files in this repo as done in [aodn/aodn_cloud_optimised@2193da4](https://github.com/aodn/aodn_cloud_optimised/commit/2193da40a0697c861183d422a0b6afe43c42cba8)
> 
> so that the unittest can run ?
> 
> there have been some changes lately, and i got the same bug

Thanks for looking into this, I was wondering what was going on. See additional commit https://github.com/aodn/data-services/pull/1208/commits/79f41e20241d9b8ba60ff1eba24d18da061fe52a

bpasquer (approved):

PR #1224, opened 2024-12-22
Update SOFAR config

xhoenner:


bpasquer (approved):

xhoenner:
@bpasquer these changes made it to prod in mid Dec but had to update the pickle file for these two deployments today as no new data was being produced before. All sorted now.

PR #1226, opened 2025-01-16
Update SOFAR config with SA deployments

xhoenner:
Still WIP, waiting for more instructions from the facility regarding several fields.


bpasquer:
Update of the config will be done in https://github.com/aodn/data-services/pull/1230
Closing

PR #1229, opened 2025-01-24
remove 'tests' in base path in main function

shaunahu:
Remove 'tests' in base path and add try-catch clause to monitor. This should fix the json file not generated issue. This script is expected to run at the root path of the file system.


weited (approved):
LGTM

PR #1236, opened 2025-02-28
add config for scanning currentmeter products

shaunahu:
Test passed.
image
Changes as follow:
- reconstruct config settings
- use subproduct name and product name to align with front end
- refactor code for scanning current meters (3 products per 2 versions)
- add test case for current meters


weited (approved):
LGTM

PR #1261, opened 2025-08-05
Feat: surface waves altimetry - add SWOT

lbesnard:


lbesnard:
nah it's not a build failing. but the tests. If you look, it can't install miniconda, nothing to do with this PR

PR #1265, opened 2025-10-26
Feat: scatterometry - add new satellites

lbesnard:


evacougnon:
yes, it hasn't been enabled yet but should be in the top list as it was impossible to enable new satellites in the pipeline config (aodndata) for most of the surface waves dataset collections (including the scatterometry). The facility is aware of the fact we were unable to update the legacy pipeline and they were happy to wait for us to fully migrate and enable the surface waves pipeline into the new infra when convenient for us.

FYI @mhidas and @bpasquer

leonardolaiolo:
@evacougnon 
here the [PR](https://github.com/aodn/chef-private/pull/4617) to enable this pipeline, keep an eye on it :+1:

evacougnon:
closing as all working in the new infra https://github.com/aodn/backlog/issues/7900

